PR #20, opened 2025-07-24
feat(cluster scan): adding acls to command output

adrian-januzi:
ACL table when a cluster has Kafka ACLs set on a principal:
Screenshot 2025-07-24 at 11 08 39

If no Kafka ACLs are set on any principals, the section isn't displayed.

---

Other changes:
- Limited the output of the 'Cluster Operations' table to the last 5 operations.
- Hid cluster nodes in the `Cluster Nodes` table if they do not have an ARN + instance type is N/A.


gcairns01 (approved):
lgtm

PR #44, opened 2025-08-18
fix(scan): kcp fails to connect to older Kafka clusters.

adrian-januzi:
An issue raised by Warren in the #kcp channel. Their user kept failing to connect to their MSK cluster using kcp due to us defaulting the Sarama KafkaAdmin client to use Kafka 4.0.0 while the brokers were on an earlier version of Kafka.

Now kcp dynamically determines the Kafka version to use in the KafkaAdmin client based on the MSK version.

### DIFF
--- a/internal/generators/scan/cluster/cluster_scanner.go
+++ b/internal/generators/scan/cluster/cluster_scanner.go
@@ -322,14 +322,7 @@ func (cs *ClusterScanner) scanKafkaResources(clusterInfo *types.ClusterInformati
 	}
 
 	clientBrokerEncryptionInTransit := types.GetClientBrokerEncryptionInTransit(clusterInfo.Cluster)
-
-	var kafkaVersion string
-	if clusterInfo.Cluster.Provisioned != nil {
-		kafkaVersion = utils.ConvertKafkaVersion(clusterInfo.Cluster.Provisioned.CurrentBrokerSoftwareInfo.KafkaVersion)
-	} else {
-		slog.Warn("⚠️ Serverless clusters return nil for Kafka version, defaulting to 4.0.0")
-		kafkaVersion = "4.0.0"
-	}
+	kafkaVersion := cs.getKafkaVersion(clusterInfo)
 
 	admin, err := cs.kafkaAdminFactory(brokerAddresses, clientBrokerEncryptionInTransit, kafkaVersion)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/generators/scan/cluster/cluster_scanner.go
```diff
@@ -322,7 +323,15 @@ func (cs *ClusterScanner) scanKafkaResources(clusterInfo *types.ClusterInformati
...
 	clientBrokerEncryptionInTransit := types.GetClientBrokerEncryptionInTransit(clusterInfo.Cluster)
 
-	admin, err := cs.kafkaAdminFactory(brokerAddresses, clientBrokerEncryptionInTransit)
+	var kafkaVersion string
+	if clusterInfo.Cluster.Provisioned != nil {
+		kafkaVersion = utils.ConvertKafkaVersion(clusterInfo.Cluster.Provisioned.CurrentBrokerSoftwareInfo.KafkaVersion)
+	} else {
+		slog.Warn("⚠️ Serverless clusters return nil for Kafka version, defaulting to 4.0.0")
+		kafkaVersion = "4.0.0"
+	}
```
glenn-smith-0, line 332:
```suggestion

func (cs *ClusterScanner) getKafkaVersion(clusterInfo *types.ClusterInformation) string {
	if clusterInfo.Cluster.ClusterType == kafkatypes.ClusterTypeServerless {
		return "4.0.0"
	}

	if clusterInfo.Cluster.ClusterType == kafkatypes.ClusterTypeProvisioned {
		return utils.ConvertKafkaVersion(clusterInfo.Cluster.Provisioned.CurrentBrokerSoftwareInfo.KafkaVersion)
	}

	return "4.0.0"
}
```

would something like this be more explicit about whats going on here?

glenn-smith-0: or even, note I haven't tested either :P

```go
func (cs *ClusterScanner) getKafkaVersion(clusterInfo *types.ClusterInformation) string {
	switch clusterInfo.Cluster.ClusterType {
	case kafkatypes.ClusterTypeServerless:
		return "4.0.0"
	case kafkatypes.ClusterTypeProvisioned:
		return utils.ConvertKafkaVersion(clusterInfo.Cluster.Provisioned.CurrentBrokerSoftwareInfo.KafkaVersion)
	default:
		slog.Warn("⚠️ Unknown cluster type, defaulting to 4.0.0")
		return "4.0.0"
	}
}
```

internal/utils/utils.go
```diff
@@ -0,0 +1,18 @@
...
+func ConvertKafkaVersion(kafkaVersion *string) string {
+	switch {
+	case strings.Contains(*kafkaVersion, "kraft"):
+		return strings.ReplaceAll(*kafkaVersion, ".x.kraft", ".0")
+	case strings.Contains(*kafkaVersion, "x"):
+		return strings.ReplaceAll(*kafkaVersion, ".x", ".0")
+	case strings.Contains(*kafkaVersion, "tiered"):
+		return strings.ReplaceAll(*kafkaVersion, ".tiered", "")
+	case *kafkaVersion == "3.6.0.1":
+		return "3.6.0"
```
glenn-smith-0, line 14: whats this all about?

internal/generators/scan/cluster/cluster_scanner.go
adrian-januzi: Yea that's smarter 🤓

glenn-smith-0 (approved):

PR #227, opened 2026-03-09
fix: graceful discovery error handling and --skip-managed-connectors flag

tjunderhill:
## Summary

- Add `--skip-managed-connectors` flag to skip connector discovery entirely when IAM permissions are unavailable
- Non-critical discovery steps (VPC connections, operations, SCRAM secrets, policy, compatible versions, connectors) now log warnings and continue instead of failing the entire cluster
- Critical steps (cluster description, bootstrap brokers, nodes, networking) remain fatal
- Removed emojis and uppercase from `fmt.Errorf` messages — error values should be plain strings, emojis belong in `slog` log lines
- All `fmt.Errorf` calls now use `%w` instead of `%v` for proper error wrapping/unwrapping
- `discoverTopics` now returns errors instead of silently logging and swallowing them

## Context

When running `kcp discover`, a missing IAM permission on a non-critical step (e.g., `kafkaconnect:ListConnectors`) would cause the entire cluster to be skipped — including topic discovery and all other data. This was because every discovery step used a hard `return nil, nil, err` pattern.

The `--skip-managed-connectors` flag helps when users know upfront they lack permissions, while the graceful degradation handles unexpected permission errors on any non-critical step without aborting.

## What stays fatal

| Step | Reason |
|------|--------|
| `describeCluster` | Core cluster identity |
| `getBootstrapBrokers` | Required for connectivity |
| `scanClusterNodes` | Required for networking subnet/broker mapping |
| `scanNetworkingInfo` | Required for HCL migration infrastructure |

## What is now graceful

| Step | Fallback |
|------|----------|
| `scanClusterVpcConnections` | Empty slice |
| `scanClusterOperations` | Empty slice |
| `scanClusterScramSecrets` | Empty slice |
| `getClusterPolicy` | Empty policy output |
| `getCompatibleKafkaVersions` | Empty versions output |
| `discoverMatchingConnectors` | Empty connectors slice |

## Test plan

- [x] New test: `TestClusterDiscoverer_ConnectorFailureDoesNotBlockTopicDiscovery`
- [x] All existing tests pass (`go test ./cmd/discover/...`)
- [ ] Manual: run `kcp discover` with IAM role missing `kafkaconnect:ListConnectors` — cluster should still be discovered with empty connectors

Fixes #226


tjunderhill (commented):
## Code Review

### Important Issues

**1. Nil vs empty slice inconsistency when `--skip-managed-connectors` is used**

When `skipManagedConnectors=true`, the `Connectors` field is never explicitly assigned — it stays `nil` (the zero value of the struct). But when connector discovery *fails*, it's explicitly set to `[]types.ConnectorSummary{}`.

This inconsistency matters because:
- JSON serialization: `nil` → `"connectors": null` vs empty slice → `"connectors": []`
- Downstream `len()` checks work identically, but `== nil` checks will behave differently

The skip path should match the error path:
```go
if skipManagedConnectors {
    slog.Info("⏭️ skipping MSK Connect connector discovery")
    awsClientInfo.Connectors = []types.ConnectorSummary{}  // ← add this
}
```

**2. `discoverTopics` is now fatal — inconsistent with the PR's graceful degradation philosophy**

The PR converts `discoverTopics` from silently swallowing errors to returning them as fatal errors that abort the entire cluster. This means a topic listing failure (e.g., missing `kafka:ListTopics` permission) will now skip the entire cluster — including all the non-critical data that this PR just made graceful.

This seems inconsistent with the PR's own design: if VPC connections, operations, SCRAM secrets, policy, versions, and connectors all degrade gracefully, why should topic discovery be fatal? Consider making topic discovery graceful too (warn + empty topics), or at minimum document this as an intentional behavioral change, since users who previously got clusters-with-empty-topics will now get no cluster at all.

### Suggestions (non-blocking)

**3. Growing boolean parameter list**

`Discover(ctx, arn, region, skipTopics, skipMetrics, skipManagedConnectors)` now has three adjacent `bool` parameters. It's easy to accidentally swap them at call sites. Consider consolidating into an options struct in a follow-up:
```go
type DiscoverOpts struct {
    SkipTopics            bool
    SkipMetrics           bool
    SkipManagedConnectors bool
}
```

**4. No test for `skipManagedConnectors=true` path**

`TestClusterDiscoverer_ConnectorFailureDoesNotBlockTopicDiscovery` tests the connector *failure* path but not the explicit *skip* path (`skipManagedConnectors=true`). A test verifying that `ListConnectors` is never called when the flag is set would lock down the behavior.

### What looks good

- The critical vs non-critical classification is well thought out and clearly documented
- `%v` → `%w` for error wrapping is correct and improves error inspection
- The new test for connector failure not blocking topics is a good regression test
- The docs/permissions split into Core, Non-Critical, and MSK Connect groups is clear and helpful

adrian-januzi:
no longer required as we are removing the topic parsing for connectors

PR #257, opened 2026-04-07
feat: add --use-sasl-plain flag to migration commands

adrian-januzi:
## Summary
- Adds `--use-sasl-plain`, `--sasl-plain-username`, `--sasl-plain-password` flags to `migration init` and `migration execute` commands
- Enables migrations from source clusters using `SASL_PLAINTEXT` listeners (e.g. Confluent Platform, self-managed Kafka with SASL/PLAIN mechanism)
- Fixes `AdminOptionForAuth` silently falling through to IAM when given `AuthTypeSASLPlain`
- Adds `WithSASLPlainAuthNoTLS` AdminOption and TLS toggle to `configureSASLTypePlainAuthentication` — CC destination path (which needs TLS) is unaffected

## Context
A user migrating from Confluent Platform to Confluent Cloud had a source cluster configured with `SASL_PLAINTEXT` + `PLAIN` mechanism. No existing auth flag combination supported this — IAM/SCRAM both hardcode TLS, unauthenticated options skip SASL, and mTLS is a different mechanism entirely.

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes (all 28 packages)
- [x] New unit tests: `WithSASLPlainAuthNoTLS` config, `configureSASLTypePlainAuthentication` with TLS on/off, `AdminOptionForAuth` SASL/PLAIN mapping
- [x] New command tests: `--use-sasl-plain` requires credentials, passes validation with credentials
- [x] E2E: `make e2e-setup && make ci-e2e-tests` (existing `--use-unauthenticated-plaintext` path still works)
- [x] Manual: test against a SASL_PLAINTEXT source cluster

### DIFF
--- a/cmd/migration/execute/cmd_migration_execute.go
+++ b/cmd/migration/execute/cmd_migration_execute.go
@@ -18,12 +18,16 @@ var (
 	awsRegion                   string
 	useSaslIam                  bool
 	useSaslScram                bool
+	useSaslPlain                bool
 	useTls                      bool
 	useUnauthenticatedTLS       bool
 	useUnauthenticatedPlaintext bool
 
 	saslScramUsername string
 	saslScramPassword string
+
+	saslPlainUsername string
+	saslPlainPassword string
 
 	tlsCaCert             string
 	tlsClientCert         string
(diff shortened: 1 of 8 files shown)


adrian-januzi:
Tested by manually updating the e2e test to setup a SASL/PLAIN cluster and use the new `--use-sasl-plain` flag.

tjunderhill (approved):
LGTM!

PR #265, opened 2026-04-14
feat(ui): chart performance optimizations

tjunderhill:
## Summary

- **Chart performance**: Throttle zoom drag to 60fps with `requestAnimationFrame`, downsample chart data to 150 points before Recharts render, and split the cost data `useMemo` into two stages so service dropdown switches skip expensive re-grouping of records
- **Loading indicator**: Show a spinner while state file loads (both CLI `--state-file` and manual upload) instead of flashing the welcome page
- **Sidebar key fix**: Use cluster ARN as React key to prevent duplicate selection highlights across regions

## Test plan

- [x] All 23 Playwright E2E tests pass
- [ ] Manually verify zoom drag is smooth on Region Cost Overview with large dataset
- [x] Manually verify service dropdown switch completes in under 500ms
- [ ] Manually verify loading spinner appears when using `--state-file` with a large file
- [ ] Manually verify charts look visually correct (downsampled vs full data)
- [ ] Verify no regression on metrics charts

### DIFF
--- a/cmd/ui/frontend/src/components/common/DateRangeChart.tsx
+++ b/cmd/ui/frontend/src/components/common/DateRangeChart.tsx
@@ -13,6 +13,7 @@ import {
 import type { CategoricalChartFunc } from 'recharts/types/chart/types'
 import { format } from 'date-fns'
 import { formatDateShort } from '@/lib/formatters'
+import { downsampleChartData } from '@/lib/downsampleData'
 
 export interface ChartDataPoint {
   date: string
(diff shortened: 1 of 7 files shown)


adrian-januzi:
I wouldn't say my graph range picking is totally smooth when I am viewing at the full year range. However, as it zooms in and decreases, it gets better.

It's a bit nitpicky but I also think its fine as is.

https://github.com/user-attachments/assets/dea534da-2e80-4bbf-a630-5a43b756a82f

adrian-januzi (approved):

PR #271, opened 2026-04-17
chore: validate state file compatibility on load and upload

bmartin494:
## Summary

Implements state file validation when loading a KCP state file, covering all three loading paths. The approach is based on team discussions that concluded files should not be rejected purely on version mismatch — if the JSON is still deseralisable, the file should be usable. Instead, targeted error feedback is provided when loading actually fails, using the version field where available to give actionable context.

### Behaviour

- **Version mismatch alone is not a rejection** — if the file deserialises successfully it is accepted regardless of whether the version matches the running binary
- **Deserialisation failure with detectable version mismatch** — surfaces a targeted "version mismatch" error as the likely cause, naming both versions
- **Deserialisation failure without detectable version mismatch** — returns a generic unmarshal error
- **File not found or unreadable** — returns a clear load error
- **Frontend upload** — simplified to a lightweight JSON object check before sending to the backend; the backend is the authority on whether the content is valid

### What changed

**`internal/types/state.go` — `NewStateFromFile()`**
- Added fallback on deserialisation failure: attempts to extract just the version from the raw bytes; if a version mismatch is detectable, surfaces that as the likely cause rather than a cryptic JSON type error
- Removed strict version check on the success path — files that deserialise correctly are accepted regardless of version

**`cmd/ui/api/api.go` — UI preload (`--state-file` flag)**
- `NewUI()` now delegates to `types.NewStateFromFile()` instead of inline `os.ReadFile` + `json.Unmarshal`
- Added `preloadError` / `preloadErrorType` fields to the `UI` struct so startup preload failures are surfaced to the frontend via `GET /state` (HTTP 422) rather than silently producing a missing-state 404
- Extracted the `/state` anonymous handler into a named `handleGetState` method for testability

**`cmd/ui/api/api.go` — UI upload (`POST /upload-state`)**
- Removed strict version check — files with a different version are accepted if they bind correctly

**`cmd/ui/frontend/src/pages/home.tsx`**
- Replaced the previous structure check (`msk_sources` or `osk_sources` presence) with a minimal JSON object check — confirms the file is a non-array JSON object before sending to the backend
- Arrays are explicitly excluded (`!Array.isArray`) to give a clean frontend error rather than a raw backend deserialisation error
- Backend is now the authority on state file validity

**Test fixtures updated**
- Existing test fixtures in `cmd/report/metrics`, `cmd/scan/schema_registry`, and the frontend test fixtures (`state-migration.json`, `state-both.json`, `state-osk-only.json`) were updated to include `kcp_build_info` with `"version": "0.0.0-localdev"`

### Tests added

- `internal/types/state_test.go`: version match, version mismatch (older/newer/empty — all succeed when deserialisable), file not found, invalid JSON, schema mismatch with fallback version surfacing
- `cmd/ui/api/api_test.go` (new file): upload version match/mismatch/empty version (all succeed), missing session, preload version match/mismatch (both succeed), preload file not found, `GET /state` error shape for generic load-error preload failures
- `cmd/ui/frontend/tests/e2e/state-validation.spec.ts` (new file): uploading a JSON array is rejected with a frontend error; uploading a state file from a different KCP version succeeds

### Related

Companion bug fix in PR #275: `loadOrCreateState()` was creating new state files without a version stamp. Fixed to use `NewStateFrom(nil)` which correctly stamps the version.

## Test plan

- [x] `go test ./internal/types/... ./cmd/ui/api/...` — all unit tests pass
- [x] `go test ./...` — full suite passes
- [x] Playwright E2E tests for state file validation pass
- [ ] Manual: run `kcp ui --state-file <unreadable-file>` and confirm the UI surfaces an error rather than showing an empty state
- [ ] Manual: upload a non-KCP JSON file via the UI and confirm a clear error is shown

### DIFF
--- a/cmd/report/metrics/cmd_report_metrics_test.go
+++ b/cmd/report/metrics/cmd_report_metrics_test.go
@@ -34,7 +34,8 @@ func TestParseMetricReporterOpts_SourceTypeMSK(t *testing.T) {
 				{ID: "osk-cluster-1"},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -80,7 +81,8 @@ func TestParseMetricReporterOpts_SourceTypeOSK(t *testing.T) {
 				{ID: "osk-cluster-1"},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -118,7 +120,8 @@ func TestParseMetricReporterOpts_ClusterIdFlag(t *testing.T) {
 				{ID: "osk-cluster-1"},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -162,7 +165,8 @@ func TestParseMetricReporterOpts_MultipleClusterIds(t *testing.T) {
 				{ID: "osk-cluster-1"},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -201,7 +205,8 @@ func TestParseMetricReporterOpts_InvalidSourceType(t *testing.T) {
 				},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -252,7 +257,8 @@ func TestParseMetricReporterOpts_NoClustersInState_MSK(t *testing.T) {
 		OSKSources: &types.OSKSourcesState{
 			Clusters: []types.OSKDiscoveredCluster{},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -285,7 +291,8 @@ func TestParseMetricReporterOpts_NoClustersInState_Neither(t *testing.T) {
 		OSKSources: &types.OSKSourcesState{
 			Clusters: []types.OSKDiscoveredCluster{},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -383,7 +390,8 @@ func TestNewReportMetricsCmd_MutuallyExclusiveFlags(t *testing.T) {
 				},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
@@ -423,7 +431,8 @@ func TestParseMetricReporterOpts_NeitherFlagProvided(t *testing.T) {
 				{ID: "osk-cluster-1"},
 			},
 		},
-		Timestamp: time.Now(),
+		KcpBuildInfo: types.KcpBuildInfo{Version: "0.0.0-localdev"},
+		Timestamp:    time.Now(),
 	}
 
 	err := state.PersistStateFile(stateFilePath)
--- a/cmd/scan/schema_registry/glue_schema_registry_scanner_test.go
+++ b/cmd/scan/schema_registry/glue_schema_registry_scanner_test.go
@@ -30,7 +30,7 @@ func TestGlueSchemaRegistryScanner_Run(t *testing.T) {
 	defer func() { _ = os.Remove(tmpFile.Name()) }()
 
 	// Write initial state
-	_, err = tmpFile.WriteString(`{"regions":[],"schema_registries":null}`)
+	_, err = tmpFile.WriteString(`{"regions":[],"schema_registries":null,"kcp_build_info":{"version":"0.0.0-localdev","commit":"unknown","date":"unknown"}}`)
 	require.NoError(t, err)
 	require.NoError(t, tmpFile.Close())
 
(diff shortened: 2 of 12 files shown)


adrian-januzi:
two qustions:

1.  `--state-file` path - should we be failing the start of the UI if an invalid JSON is passed?

For example, if we pass the following JSON to the UI (`regions` should be an array, not string):

```json
{
  "msk_sources": {
    "regions": ""
  },
  "osk_sources": {
    "clusters": []
  },
  "kcp_build_info": {
    "version": "9.9.9",
    "commit": "MYFAKECOMMITHASH",
    "date": "1970-01-01T00:00:00Z"
  },
  "timestamp": "1970-01-01T00:00:00.000000+01:00"
}
```

The UI returns the following error when directly uploading the state file:
```
Error: code=400, message=Unmarshal type error: expected=[]types.DiscoveredRegion, got=string, field=msk_sources.regions, offset=28, internal=json: cannot unmarshal string into Go struct field MSKSourcesState.msk_sources.regions of type []types.DiscoveredRegion
```

However, when passing the invalid state file to the `--state-file` UI flag, there is only this warning but the UI continues to start up. In my opinion, I think it should failover instead:
```shell
2026/04/27 16:58:40 ERROR Failed to load state file path=not-a-valid-statefile.json error=failed to unmarshal state: json: cannot unmarshal string into Go struct field MSKSourcesState.msk_sources.regions of type []types.DiscoveredRegion
```

or if there is a version mismatch:
```shell
2026/04/27 16:59:10 ERROR Failed to load state file path=not-a-valid-statefile.json error=state file version mismatch: file was created with KCP version "9.9.9" but you are running version "0.0.0-localdev" — please re-export the state file using the current version of KCP or upgrade/downgrade KCP to match
```

---

2. Should we seeing the same `...state file version mismatch...` type errors in the frontend too regardless of how the statefile is uploaded (CLI flag or upload modal)? 

Maybe I am doing something silly but I can't see anything besides `No clusters available. Please upload a KCP state file to explore your infrastructure.` and the browser console only returns `Failed to load resource: the server responded with a status of 422 (Unprocessable Entity)`.

adrian-januzi:
Just one other thing I was thinking of while writing this out - should we possibly expand this to any and all commands that reads from (and maybe also writes to) the state file in some way, shape or form?

A simple scenario would be for the `create-asset` commands, if we decided to rename say `kafka_admin_client_information.cluster_id` to `kafka_admin_client_information.kafka_cluster_id` between releases, we would run into a problem where an old state file on a newer binary would throw an error or possibly panic. This work would probably go into a separate PR but food for thought.

bmartin494:
> two qustions:
> 
> 1. `--state-file` path - should we be failing the start of the UI if an invalid JSON is passed?
> 
> For example, if we pass the following JSON to the UI (`regions` should be an array, not string):
> 
> ```json
> {
>   "msk_sources": {
>     "regions": ""
>   },
>   "osk_sources": {
>     "clusters": []
>   },
>   "kcp_build_info": {
>     "version": "9.9.9",
>     "commit": "MYFAKECOMMITHASH",
>     "date": "1970-01-01T00:00:00Z"
>   },
>   "timestamp": "1970-01-01T00:00:00.000000+01:00"
> }
> ```
> 
> The UI returns the following error when directly uploading the state file:
> 
> ```
> Error: code=400, message=Unmarshal type error: expected=[]types.DiscoveredRegion, got=string, field=msk_sources.regions, offset=28, internal=json: cannot unmarshal string into Go struct field MSKSourcesState.msk_sources.regions of type []types.DiscoveredRegion
> ```
> 
> However, when passing the invalid state file to the `--state-file` UI flag, there is only this warning but the UI continues to start up. In my opinion, I think it should failover instead:
> 
> ```shell
> 2026/04/27 16:58:40 ERROR Failed to load state file path=not-a-valid-statefile.json error=failed to unmarshal state: json: cannot unmarshal string into Go struct field MSKSourcesState.msk_sources.regions of type []types.DiscoveredRegion
> ```
> 
> or if there is a version mismatch:
> 
> ```shell
> 2026/04/27 16:59:10 ERROR Failed to load state file path=not-a-valid-statefile.json error=state file version mismatch: file was created with KCP version "9.9.9" but you are running version "0.0.0-localdev" — please re-export the state file using the current version of KCP or upgrade/downgrade KCP to match
> ```
> 
> 2. Should we seeing the same `...state file version mismatch...` type errors in the frontend too regardless of how the statefile is uploaded (CLI flag or upload modal)?
> 
> Maybe I am doing something silly but I can't see anything besides `No clusters available. Please upload a KCP state file to explore your infrastructure.` and the browser console only returns `Failed to load resource: the server responded with a status of 422 (Unprocessable Entity)`.

1. "Should we be failing the start of the UI if an invalid JSON is passed?" - NewUI now returns an error and the process exits before the server starts.
 2. "Should we see the same errors in the frontend regardless of how the state file is uploaded?" Changed the flow here to match how this happens in the actual UI, rejecting the state file upload. The bad state file now prevents the UI from starting entirely, and the error is shown in the terminal explaining this. Basically if the statefile does not deserialise the UI won't start, otherwise it will load and flag that there are no data sources if the json is valid but not containing the right stuff.
@adrian-januzi

adrian-januzi:
Just one nit though I'm not too fussy on in hindsight. Maybe get someone else's opinion on it at standup - when launching the UI with an invalid state file, the UI process doesn't actually not start.

```shell
❯ cat empty.json | jq
{}
❯ kcp ui --state-file empty.json
Executing kcp with build version=0.0.0-localdev commit=b48f152915ed104404ff7498920aad8ba34f14ee date=2026-04-28T16:11:37Z
2026/04/28 17:41:00 WARN State file contains no sources — run a scan to populate it path=empty.json

kcp ui is available at http://localhost:5556
^C
❯ cat not-a-valid-statefile.json | jq
{
  "msk_sources": {
    "regions": []
  },
  "osk_sources": {
    "clusters": []
  },
  "kcp_build_info": {
    "version": "9.9.9",
    "commit": "MYFAKECOMMITHASH",
    "date": "1970-01-01T00:00:00Z"
  },
  "timestamp": "1970-01-01T00:00:00.000000+01:00"
}
❯ kcp ui --state-file not-a-valid-statefile.json
Executing kcp with build version=0.0.0-localdev commit=b48f152915ed104404ff7498920aad8ba34f14ee date=2026-04-28T16:11:37Z

kcp ui is available at http://localhost:5556
^C
```

Though I think there has been a regression in that I can't seem to trigger the version mismatch error anymore? Is that expected? As for the `no sources error`, I think we should be checking inside the `MSKSources.Regions` and `OSKSources.Clusters` for entries. Otherwise, I can pass any JSON file which will pass.

adrian-januzi (approved):
as discussed in standup, lgtm

bmartin494:
Think I'm finally happy with this from where we started and ended up in terms of handling various state file scenarios. The core behaviour: a version difference alone never rejects a state file, if it deserialises cleanly, it loads regardless of version. Rejection only happens when deserialisation itself fails, at which point the error message includes as much context as could be extracted from the file. 

A small distinction in the error messages worth calling out:
                                                                                                                                      
There are two different deserialisation failure paths in NewStateFromFile:                                                        
                                                                                                                                      
  1. JSON parses but fails to unmarshal into the struct (e.g. a field has the wrong type like regions: ""). Here we can still read the version from the raw bytes, so the message is "state file could not be loaded: <error> (file was created with KCP version X, you are running Y - ...)". We have enough context to give version-specific advice.                                                      
  2. JSON is so broken it can't be parsed at all (e.g. truncated or malformed). We can't extract anything from the file, so the message is "failed to unmarshal state file: <error> - please recreate the state file using kcp discover or kcp scan clusters". No version context is possible here.

Have checked this against the following files:
[empty.json](https://github.com/user-attachments/files/27213072/empty.json)
[faulty-json-state.json](https://github.com/user-attachments/files/27212506/faulty-json-state.json)
[kcp-state.json](https://github.com/user-attachments/files/27212508/kcp-state.json)
[missing-build-info-state.json](https://github.com/user-attachments/files/27212509/missing-build-info-state.json)
[not-a-valid-statefile.json](https://github.com/user-attachments/files/27212510/not-a-valid-statefile.json)
[random-json-state.json](https://github.com/user-attachments/files/27212511/random-json-state.json)
[version-mismatch-valid-state.json](https://github.com/user-attachments/files/27212512/version-mismatch-valid-state.json)

image

adrian-januzi (approved):
lgtm!

PR #274, opened 2026-04-23
fix(docs): set alias_type=redirect for mike

adrian-januzi:
## Summary
Configures mike's default `alias_type` as `redirect` under `plugins.mike` in `mkdocs.yml` so the `latest` alias is materialised as a redirect stub instead of a POSIX symlink. GitHub Pages rejects symlinks (*"Failed to deploy site, please ensure the repository does not contain any hard links, symlinks"*), which is mike's default alias representation on Linux.

## Note on the two commits
The first commit put the setting under `extra.version`, which is where Material's version-selector UI reads its config from — but mike reads its own config from `plugins.mike`, so that placement was silently ignored. The second commit moves it to the correct block and verifies via an isolated `mike deploy --branch test-gh-pages` that the resulting tree contains zero symlinks and a proper HTML redirect at `latest/index.html`. Worth squash-merging.

## Why this is needed
The initial backfill of the versioned docs site hit exactly the symlink failure, and the site did not publish until `latest` was redeployed with `--alias-type=redirect` on the CLI. Setting it in the plugin block makes every `mike deploy` correct by default, without relying on CLI flags.

Pairs with confluentinc/kcp-release#52, which adds `--alias-type=redirect` to the release pipeline as a belt-and-braces guard.

## Test plan
- [x] Ran `mike deploy --branch test-gh-pages --update-aliases pr-274-test-v2 latest` locally against the PR branch (deliberately omitted `--alias-type` to prove the config alone is enough). Result: `git ls-tree -r test-gh-pages | awk '$1 == "120000"'` printed nothing; `latest/index.html` is an HTML redirect to `../pr-274-test-v2/`.
- [x] After release pipeline runs with this merged, verify the Pages deployment shows `state=success`.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -38,6 +38,12 @@ theme:
 plugins:
   - search
   - callouts
+  # GitHub Pages rejects POSIX symlinks (mike's default on Linux), so the
+  # `latest` alias must be a redirect stub instead. Without alias_type=redirect,
+  # release deploys fail with "Failed to deploy site, please ensure the
+  # repository does not contain any hard links, symlinks".
+  - mike:
+      alias_type: redirect
 
 markdown_extensions:
   - admonition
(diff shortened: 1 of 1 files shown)


tjunderhill (approved):
LGTM!

PR #275, opened 2026-04-23
bug: stamp build version when creating new state file

bmartin494:
## Summary

`loadOrCreateState()` in `cmd/scan/clusters` was constructing a new `State` struct directly with an empty `KcpBuildInfo{}` when no existing state file was found. This meant a first-time scan (particularly OSK) would write a state file to disk with no version stamped — inconsistent with `NewStateFrom()` which always stamps the current binary version via `build_info.Version`.

## What changed

- Replaced the direct struct construction in `loadOrCreateState()` with `NewStateFrom(nil)`, which correctly stamps `Version`, `Commit` and `Date` from the running binary
- Explicitly initialised `SchemaRegistries` after the call to preserve the existing behaviour — `NewStateFrom(nil)` does not initialise `SchemaRegistries` so this was carried over from the original code

## How it was found

Discovered while investigating state file version validation as part of [chore/reject-statefile-version-mismatch](#271). A first-time OSK scan produces a state file with an empty version field, which would silently fail any version-based or `kcp_build_info`-based validation on upload to the UI.

## Tests added

Unit tests added in `cmd/scan/clusters/load_or_create_state_test.go` covering the four key cases for `loadOrCreateState()`:

- **New file — version stamping**: confirms `Version`, `Commit`, and `Date` are all stamped from the running binary when creating a new state file
- **New file — field initialisation**: confirms `MSKSources`, `OSKSources`, and `SchemaRegistries` are initialised as non-nil with empty slices (guards against nil pointer panics in downstream code)
- **Existing valid file**: confirms an existing state file is loaded correctly
- **Corrupt file**: confirms an error is returned rather than silently producing empty state

## Test plan

- [x] `go test ./cmd/scan/clusters/...` — all tests pass including new ones
- [x] `go test ./...` — full suite passes
- [ ] Manual: run `kcp scan clusters --source-type osk` against a fresh directory with no existing state file and confirm the resulting `kcp-state.json` has `kcp_build_info.version` populated

### DIFF
--- a/cmd/scan/clusters/cmd_scan_clusters.go
+++ b/cmd/scan/clusters/cmd_scan_clusters.go
@@ -214,13 +214,9 @@ func runScanClusters(cmd *cobra.Command, args []string) error {
 func loadOrCreateState(stateFilePath string) (*types.State, error) {
 	if _, err := os.Stat(stateFilePath); os.IsNotExist(err) {
 		slog.Info("creating new state file", "file", stateFilePath)
-		return &types.State{
-			MSKSources:       &types.MSKSourcesState{Regions: []types.DiscoveredRegion{}},
-			OSKSources:       &types.OSKSourcesState{Clusters: []types.OSKDiscoveredCluster{}},
-			SchemaRegistries: &types.SchemaRegistriesState{},
-			KcpBuildInfo:     types.KcpBuildInfo{},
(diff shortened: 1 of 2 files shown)


tjunderhill (approved):
LGTM!

PR #278, opened 2026-04-24
feat: add OSK cluster support to TCO Inputs page

tjunderhill:
## Summary

- Adds source-type tabs (MSK/OSK) to the TCO Inputs page
- Displays structured OSK metadata in column headers (environment, location, Kafka version, sorted labels)
- Enables unified metrics modal lookup for both MSK and OSK clusters, including partition metrics via JMX/Prometheus
- Shows explicit missing-metric message with fallback dropdown when a preselected metric is unavailable (instead of silently switching)
- Scopes CSV export to the active tab with RFC 4180 escaping
- Adds server-side date filtering for OSK metrics via `FilterClusterMetrics` with auto-detection
- Refactors `parseDateRange` to eliminate duplicated date-parsing logic across 3 API handlers
- Prunes stale TCO workload data when the cluster set changes
- Type-safe `WorkloadField` and `MetricType` props on `TCOInputRow` — no `as` casts

## New files

- `useOSKTCOClusters` hook — extracts OSK clusters from state into `TCOCluster` format
- `TCOInputRow` component — extracted from inline table rows with type-safe props
- `useTCOModal` hook — manages metrics modal state for both MSK and OSK
- `tco-inputs.spec.ts` — 7 E2E tests for tab switching, metadata, CSV preview, label pills
- `tco-inputs-metrics.spec.ts` — 5 E2E tests for metrics modal, value transfer, partition lookup, date range
- `region-costs.spec.ts` — 3 E2E tests for region cost date filtering
- `state-with-metrics.json` — test fixture with 7 OSK metric types matching JMX/Prometheus output

## Test plan

- [x] `npx tsc --noEmit` — zero errors
- [x] 12/12 TCO E2E tests pass (`tco-inputs.spec.ts` + `tco-inputs-metrics.spec.ts`)
- [x] 27/27 Go unit tests pass for `FilterClusterMetrics` (MSK, OSK, auto-detection, edge cases)
- [x] 3 rounds of superpowers code review — all actionable issues fixed
- [x] Manual: TCO Inputs → MSK tab shows MSK clusters with metrics lookup
- [x] Manual: TCO Inputs → OSK tab shows OSK clusters with metadata, metrics lookup, partition lookup
- [x] Manual: Metrics modal → missing metric shows fallback message with dropdown


adrian-januzi:
## Mega-review findings — addressable items only

Ran a 4-engine review (general quality, security, Go, interface audit) against `pr-278-review` in a worktree. No critical blockers, but the list below covers items I'd want addressed before merge.

### Must-fix (correctness / feature regressions)

1. **Fallback-metric dropdown silently reverts the user's pick** — `cmd/ui/frontend/src/hooks/useMetricSelection.ts:39`
   The effect now unconditionally writes `selectedMetric = modalPreselectedMetric` when `inModal && modalPreselectedMetric` is truthy, and `selectedMetric` is in the dep array. When the user picks an alternative from the new "View available metrics instead" `<Select>`, the effect immediately overwrites it with the missing metric — the feature this PR adds doesn't actually work. Track whether the user has interacted (a `userSelected` ref), or re-introduce the `availableMetrics.includes(modalPreselectedMetric)` guard with a one-shot write. Flagged by two independent engines.

2. **`handleGetOSKMetrics` regresses the "no metrics collected" response** — `cmd/ui/api/api.go:234`
   The old handler returned `404` with actionable guidance (`Run 'kcp scan clusters --source-type osk --metrics jolokia' or '--metrics prometheus' to collect metrics`). After the refactor, `FilterClusterMetrics` returns `nil` metrics with **no error**, so the handler emits `200 OK` with an empty payload. Not covered by tests. Either detect `filteredMetrics.Metrics == nil` post-call and return the old 404, or have `filterOSKClusterMetrics` return a typed sentinel (`ErrNoMetrics`) the handler can distinguish from cluster-not-found. Add a response-body/status assertion for the "cluster exists, no metrics" path.

3. **`setKcpState` wipes all TCO workload data on every upload** — `cmd/ui/frontend/src/stores/store.ts:198`
   `tcoWorkloadData: {}` is reset unconditionally — any values the user typed are lost on re-upload, even when the cluster set is unchanged. Redundant with `initializeTCOData`'s per-cluster pruning. Drop the blanket reset, or diff the cluster set first.

4. **Headline feature has no tests** — `internal/services/report/report_service_test.go`
   Every OSK subtest in `TestFilterClusterMetrics_SourceAware` calls `FilterClusterMetrics` with `startTime=nil, endTime=nil`, so the date-filter branch — the PR's stated purpose — is never exercised on the OSK path. Add table-driven cases for: RFC3339-with-offset timestamps, inclusive start/end boundaries, an unparseable `Start` (should be silently skipped), and a subset-producing range. `parseDateRange` itself also has no direct test (cheap to add via `httptest.NewRecorder()`).

### Should-fix (a11y, consistency, safety)

5. **New TCO tablist duplicates and visually diverges from `components/common/Tabs`** — `cmd/ui/frontend/src/components/tco/TCOInputsPage.tsx:121`
   Different color tokens (`border-primary` vs the accent used everywhere else), different border thickness, different spacing. Extend `Tabs` with an optional count badge and use it here.

6. **ARIA tab pattern is half-wired** — same file/line
   `role='tablist'`/`role='tab'` are declared but there's no `aria-controls`, no `role='tabpanel'`/`aria-labelledby` on the table region, inactive tabs remain in the tab order, and arrow-key navigation isn't handled. This is strictly worse for AT users than plain buttons. Complete the pattern or drop the roles.

7. **Icon-only metrics-link buttons have no `aria-label`** — `cmd/ui/frontend/src/components/tco/TCOInputRow.tsx:109`
   `title` alone is unreliable on VoiceOver/NVDA. Add `aria-label={buttonTitle}` (ideally including the cluster name).

8. **TCO numeric inputs are unlabelled** — `cmd/ui/frontend/src/components/tco/TCOInputRow.tsx:99`
   No `id`/`htmlFor`/`aria-label`/`aria-labelledby`. Screen readers read every input as an unlabelled spinner. Add `aria-label={`${label} for ${cluster.name}`}` and mark the row-label `<td>` as `<th scope='row'>`.

9. **Metrics-fetch effect has no race-condition guard** — `cmd/ui/frontend/src/hooks/useClusterMetricsFetch.ts:41`
   Now that the effect refires on date changes, a slower earlier response can land after a later one and overwrite fresher data. Add an `AbortController` (or a `let cancelled = false` flag) and pass the signal through to `apiClient`.

10. **`escapeCSV` misses `\r` and doesn't neutralize formula injection** — `cmd/ui/frontend/src/lib/tcoUtils.ts:42`
    (a) RFC 4180 §2.6 requires quoting for any line-break incl. CR/CRLF — the regex is `/[,"\n]/`, add `\r`.
    (b) Cells starting with `=`, `+`, `-`, `@`, TAB, or CR are executed as formulas by Excel/Sheets/LibreOffice. OSK cluster IDs are user-authored and flow into headers; KCP state files are designed to be shared. The PR advertises "RFC 4180 escaping" as a feature — close the gap:
    ```ts
    if (/^[=+\-@\t\r]/.test(val)) val = "'" + val;
    if (/[,"\r\n]/.test(val)) val = `"${val.replace(/"/g, '""')}"`;
    ```

11. **Initial tab defaults to MSK even when only OSK clusters exist** — `cmd/ui/frontend/src/components/tco/TCOInputsPage.tsx:26`
    `useState<SourceType>(() => mskClusters.length === 0 && oskClusters.length > 0 ? SOURCE_TYPES.OSK : SOURCE_TYPES.MSK)`.

_Engines run: General Quality, Security Specialist, Go Specialist, Interface Audit. Structured-personas engine (`ce:review`) was unavailable — coverage gap noted. Backend Go tests were not re-executed; assessment is static._

tjunderhill:
Addressed in c00f54d:

- **Item 1 (effect overwrites dropdown selection)**: Fixed. Added a `userOverrodePreselection` ref that gates the effect after the user picks an alternative metric from the fallback dropdown.
- **Item 2 (200 OK instead of 404 for missing metrics)**: Fixed. Restored the 404 with actionable guidance (`--metrics jolokia` / `--metrics prometheus`).
- **Item 4 (no date-filter tests for OSK)**: Fixed. Added 8 tests covering start/end boundaries, RFC3339 with timezone offsets, unparseable timestamps, subset filtering, and aggregate recalculation.

**Item 3 (setKcpState wipes TCO workload data)**: This is intentional. Uploading a new state file represents a new dataset — preserving stale workload values from a previous file would be confusing. `initializeTCOData` handles the per-cluster initialization after upload.

adrian-januzi (approved):

PR #284, opened 2026-05-05
docs: document Prometheus PromQL queries in metrics collection guide

ifeeneyconfluent:
Help users configure their Prometheus setup by listing the exact PromQL queries kcp submits, the expected metric names, and the rate window logic.


adrian-januzi (approved):
sounds good to me

pinaruyar (approved):

PR #286, opened 2026-05-06
feat: discover Connect topics by suffix instead of exact name

ifeeneyconfluent:
Use suffix matching for connect-configs and connect-status topics so that prefixed topic names (e.g. myapp.connect-configs) are also discovered when scanning for self-managed connectors.


adrian-januzi:
No longer required.

PR #290, opened 2026-05-08
feat: add OSK metrics query transparency for Jolokia and Prometheus

bmartin494:
## Summary

- Surfaces the exact Jolokia MBean paths and Prometheus PromQL queries used during `kcp scan clusters --source-type osk`, so users can see, understand, and reproduce them in the UI Query tab and CLI markdown reports
- Mirrors the CloudWatch query transparency added for MSK in #230, extending it to the two OSK metrics backends
- Extends `MetricQueryInfo` with a `source_type` discriminator and source-specific optional fields (backward-compatible — existing MSK state files are unchanged)
- Introduces typed `MetricBackend` constants (`MetricBackendCloudWatch`, `MetricBackendJolokia`, `MetricBackendPrometheus`) following the existing `SourceType` pattern
- Fixes GlobalPartitionCount to query the real controller MBean instead of aliasing PartitionCount
- Bumps `fast-uri` 3.1.0 → 3.1.2 to fix CVE-2026-6321 and CVE-2026-6322

### Jolokia (JMX)
- MBean path, Jolokia endpoint URL, and copyable **curl command** for each metric
- Statistic (Rate/Sum), poll interval, and query duration displayed in UI grid and markdown report
- Aggregation notes describing how each metric is computed, with auth hint for authenticated endpoints
- New `controllerMBeans` category for MBeans that only exist on the controller broker (tries all brokers, uses first success, logs info if unavailable)

### Prometheus
- Resolved PromQL query, Prometheus URL, raw metric name, and copyable **curl command** using `query_range` API with actual start/end timestamps and step
- Statistic, query step, and query duration displayed in UI grid and markdown report
- Aggregation notes describing rate windows, step sizes, and GiB conversion

### GlobalPartitionCount fix
- **Previously:** both `PartitionCount` and `GlobalPartitionCount` used the same query (`kafka.server:type=ReplicaManager,name=PartitionCount`), which counts partition replicas per broker — `sum()` gives total replicas, not unique partitions
- **Now (Jolokia):** queries `kafka.controller:type=KafkaController,name=GlobalPartitionCount` from the controller broker directly
- **Now (Prometheus):** queries `kafka_controller_kafkacontroller_value{name="GlobalPartitionCount"}` — the actual metric format produced by the JMX exporter for controller MBeans
- If the metric is unavailable (controller pods not scraped), it is gracefully omitted with an info log — all other metrics still collected
- **Verified on playground:** PartitionCount=2368 (total replicas), GlobalPartitionCount=790 (unique partitions), confirming RF=3 (790 × 3 ≈ 2368)
- Docs updated with guidance on scraping controller nodes (KRaft controllers may run as separate pods requiring their own PodMonitor)

### UI
- Refactored `ClusterMetricsQueryTab` into source-specific sub-components with coloured source badges (CloudWatch / Jolokia / Prometheus)
- 2-column parameter grid on all source types showing Statistic, Period/Step, Query Duration, and endpoint info (consistent layout across Jolokia, Prometheus, and CloudWatch)
- Copy-to-clipboard on curl commands, PromQL queries, and MBean paths
- Defensive guards on optional fields (MBean path, Prometheus metric name)
- Fixed React hooks rule violation — `useState` moved before early return

### Reports
- Metrics markdown report now renders query details for all source types (was MSK-only)
- Jolokia and Prometheus sections include Statistic, Poll Interval/Query Step, and Query Duration matching the UI layout

### Other
- `FormatQueryDuration` helper formats durations readably (e.g. `5m`, `7d2h` instead of `170h0m0s`)
- `MetricBackend` typed constants replacing string literals, matching the existing `SourceType` pattern
- Renamed `PrometheusMetric` field to `PrometheusMetricName` to match JSON tag
- Set `SourceType: MetricBackendCloudWatch` on all CloudWatch query info (including GlobalPartitionCount)
- Bumped `fast-uri` 3.1.0 → 3.1.2 (CVE-2026-6321, CVE-2026-6322 — path traversal vulnerabilities)
- Docs: corrected GB → GiB for TotalLocalStorageUsage (computation uses 1024^3)

## Test plan

- [x] Unit tests for `buildJMXQueryInfo` and `CollectOverDuration` QueryInfo population (dynamic count from MBean slices)
- [x] Unit tests for `buildPrometheusQueryInfo` and `CollectMetrics` QueryInfo population (7 metrics)
- [x] Unit tests for `FormatQueryDuration` (13 cases including day boundaries)
- [x] Unit test for controller MBean graceful omission (mock returns 404, metric omitted from data but present in query info)
- [x] Report tests for Jolokia query details (source label, MBean path, statistic, curl, absence of CloudWatch fields)
- [x] Report tests for Prometheus query details (PromQL, real timestamps in curl, statistic, absence of CloudWatch/Jolokia fields)
- [x] Backward-compat test for QueryInfo with empty SourceType (covers default CloudWatch rendering branch)
- [x] All existing MSK metric service and report tests pass (no regressions)
- [x] TypeScript type-checks clean
- [x] Playwright E2E tests pass (40/40)
- [x] OSK integration tests pass (all variants: plaintext, SASL, TLS, SASL+SSL, JMX, Prometheus)
- [x] Manual: verified Prometheus state file `query_info` with real scan data
- [x] Manual: verified Jolokia state file `query_info` with real scan data
- [x] Manual: verified Query tab renders correctly in UI for both backends
- [x] Manual: verified `kcp report metrics --source-type osk` markdown output
- [x] Manual: verified generated curl commands work against live brokers
- [x] Manual: verified GlobalPartitionCount from controller MBean via Docker (Jolokia, value=51)
- [x] Manual: verified GlobalPartitionCount graceful omission when MBean unavailable (playground Jolokia)
- [x] Manual: verified GlobalPartitionCount=790 vs PartitionCount=2368 on playground (Prometheus, RF=3)
- [x] Manual: verified full metric selector displayed in UI for GlobalPartitionCount


bmartin494:
Prometheus UI:
image
Jolokia UI:
image

adrian-januzi:
I noticed when running this myself and testing the Prometheus cURL command that it still had timestamp placeholders as seen in the screenshot.

Screenshot 2026-05-11 at 10 18 12

I think this was working on Friday when we paired so it might be something that I've done different by using the `--metrics-range 30d` instead of `--metrics-range 7d`?

adrian-januzi:
Same problem exists in the markdown reports:
```shell
kcp report metrics --cluster-id osk-cfk-playground --state-file kcp-state.json
```
Screenshot 2026-05-11 at 10 30 48

Also, I think we should add the duration + 'statistics' (SUM/AVG/etc.) into the markdown report too so that it matches the UI view as closely as possible (even if no one will ever use the `kcp report` command :) ).

adrian-januzi:
I think a general 'nice to have' would be to use typed values for the different metrics types - cloudwatch, prometheus and jolokia.

Something like:
```golang
  const (
      MetricBackendCloudWatch MetricBackend = "cloudwatch"
      MetricBackendJolokia    MetricBackend = "jolokia"
      MetricBackendPrometheus MetricBackend = "prometheus"
  )
```

So that in the code they can be references as `types.MetricBackendCloudWatch`. The same thing is done for the [cluster type](https://github.com/confluentinc/kcp/blob/main/internal/types/state.go#L951-L956) - osk or msk - so it makes sense to repeat the pattern here imo.

adrian-januzi:
I think some tests covering Prometheus/Jolokia in the [metrics report tests](https://github.com/confluentinc/kcp/blob/main/internal/services/metrics/metric_service_test.go) file would be good as they are similar to CloudWatch but not close enough to avoid making new tests.

adrian-januzi (approved):
lgtm

PR #291, opened 2026-05-11
feat: add --mode new and pattern selection to migrate-topics

adrian-januzi:
Extends `kcp create-asset migrate-topics` so it can emit either cluster-link mirror topics (existing behavior) or plain Confluent Cloud topics for greenfield migrations. Adds glob-based topic selection and switches output to one .tf file per topic.

- `--mode {mirror|new}` is now required; `new` emits `confluent_kafka_topic` resources with partitions and CC-supported configs preserved from source (replication.factor and non-allow-listed configs are dropped).
- `--topics-include` / `--topics-exclude` accept glob patterns; empty include defaults to all non-internal topics, exclude wins on overlap. An empty selection with patterns set is a hard error rather than a silent empty TF directory.
- Output is one `<sanitized-topic>.tf` per topic plus shared `providers.tf` and `variables.tf` for both modes. Filename collisions surface as a hard error.
- The CC-supported topic config allow-list lives in `internal/services/hcl/confluent/topic.go` (snapshot 2026-05-11). A header comment in each new-mode file and a CLI note both point at the upstream docs.
- UI wizard supports both 'mirror' and 'new' topic migration routes.


adrian-januzi:
Migration a subset of all topics - those starting with `demo-*`.
```shell
kcp create-asset migrate-topics --state-file kcp-state.json --source-type msk --cluster-id arn:aws:kafka:us-east-1:635910096382:cluster/kcp-playground/0251db9f-2675-4d50-b9f1-97e171c6b3ed-5 --mode new --target-cluster-id cc-cluster --target-rest-endpoint cc-rest-endpoint --topics-include 'demo-*'
```

If the glob cannot be matched, an error is returned. This avoids incidentally migrating all topics instead of the specific subset. However, if anyone disagrees and instead we should gracefully continue - warn the user the glob could not be matched and instead create a `migrate-topics` dir with all topics instead.

tjunderhill (approved):
LGTM!

PR #292, opened 2026-05-12
feat: kcp report plan — sizing + cluster-type mapping (experimental)

suhothayan:
## Summary

New `kcp report plan` subcommand. Consumes a `kcp-state.json` (and an optional `plan-inputs.yaml`) and emits `plan.md` + `plan.json` with:

- Source-environment summary (with fleet-wide broker + topic totals)
- Per-cluster sizing (throughput + partition ratios + headroom; canonical lowercase `p95` / `p99` / `max`)
- Enterprise / Dedicated decision via a 6-rule hard-limit catalog
- Dedicated topology (Multi-Zone / Single-Zone) and CKU unit on Dedicated verdicts
- PrivateLink / PNI / Transit Gateway / VPC Peering networking decision
- Per-cluster rationale lines + ⚠ cost callout when a customer-declared flag forced Dedicated
- "Actions Needed" section listing concrete state-file gaps to close (missing metrics, ACL scan gaps, etc.) — omitted when there are none
- Collapsed sizing-math appendix (single table that scales to multi-dozen-cluster fleets)

**Determinism:** same state file + same plan-inputs + same KCP version → byte-identical JSON output. Verified by a build-twice test (`plan_service_test.go:TestPlanServiceBuild_DeterministicByteIdenticalJSON`) and by repeated runs producing md5-identical normalized JSON.

## Explicitly out of scope (deferred to follow-up PRs)

- Auth approach
- Switchover recommendation + prerequisites
- Red flag rules
- Cost-vs-inventory reconciliation
- Workstreams / decision blockers / open-question catalogue beyond the MVP "Actions Needed" surface

`plan_schema_version` is set to `"1-experimental"` while we iterate on the surface — Migration Hub can consume now with additive-only expectations.

## How it works

```bash
kcp report plan --state-file kcp-state.json

# With overrides
kcp report plan --state-file kcp-state.json --plan-inputs plan-inputs.yaml --output md
```

`plan-inputs.yaml` accepts:

- **Sizing knobs** — `sla_target`, `sizing_percentile`, `headroom_fraction`, `privatelink_safety_threshold`, `spiky_workload_ratio`.
- **Customer-declared hard-requirement flags** — `enforce_schemas_at_the_broker`, `requires_high_throughput_rest_produce_api`, `requires_99_95_sla_within_a_single_zone`. These force Dedicated; setting one in error costs 5–10× monthly, so the renderer surfaces a ⚠ cost callout on the Dedicated verdict whenever one of them is the reason. The callout names the specific flag and the recovery path.
- **Networking-topology signals** — `target_cloud`, `existing_vpc_connectivity` (for the Dedicated path's TGW / VPC Peering variants).

Determinism is preserved: identical state file + identical plan-inputs → identical JSON. The hard-requirement flags participate in the input set; the ⚠ callout is the UX-level guard against accidental Dedicated escalation, not a determinism mechanism.

## Sample output

Two synthetic fixtures live under `docs/assets/`:

- [`plan.sample.md`](https://github.com/confluentinc/kcp/blob/feat/report-plan-cluster-type-mvp/docs/assets/plan.sample.md) — 4-cluster "complete" state with plan-inputs supplied. Exercises every MVP verdict path: Enterprise + PrivateLink, Enterprise + PNI, Dedicated Multi-Zone via state-derived ACL rule, Dedicated Multi-Zone via mTLS-off-AWS. Zero entries in Actions Needed.
- [`plan.sample.json`](https://github.com/confluentinc/kcp/blob/feat/report-plan-cluster-type-mvp/docs/assets/plan.sample.json) — same fixture in canonical JSON. Snake_case keys + `omitempty` so Enterprise verdicts cleanly omit `topology` / `final_cku` / unset plan-input fields.
- [`plan-inputs.example.yaml`](https://github.com/confluentinc/kcp/blob/feat/report-plan-cluster-type-mvp/docs/assets/plan-inputs.example.yaml) — the input-overrides template, grouped by topic with one-line comments on each field.

## Test plan

- [x] `go test ./internal/types/... ./internal/services/plan/... ./internal/services/report/... ./cmd/report/...` passes
- [x] `go vet ./...` clean
- [x] CLI `--help` matches sibling `cmd/report/costs` and `cmd/report/metrics` layout (Required / Optional sections + env-var binding via `utils.BindEnvToFlags`)
- [x] Determinism: repeated runs produce md5-identical normalized JSON (modulo `generated_at`)
- [x] Scale: 30+-cluster fixture produces a readable single audit table
- [x] Graceful degradation: sizing falls back to SLA floor with a `_metrics missing_` marker + an Actions Needed entry when the state file lacks P95 metrics
- [x] Cost callout: customer-declared `requires_99_95_sla_within_a_single_zone: true` lands Dedicated and surfaces the ⚠ callout naming the flag
- [x] Per-rule coverage: one test per hard-limit rule + null-ACLs skip + Enterprise-ignores-VPC-connectivity guard

### DIFF
--- a/internal/services/report/report_service.go
+++ b/internal/services/report/report_service.go
@@ -675,8 +675,9 @@ func nearestRankIndex(n int, p float64) int {
 }
 
 // calculateMetricsAggregates returns avg, min, max, P95, P99, and count
-// per metric label. Percentiles use the lower-index nearest-rank method:
-// `values[int(N*p)]` after sorting.
+// per metric label. Percentiles use the nearest-rank method (R type 1):
+// after sorting, index = ceil(N*p) - 1, clamped to [0, N-1]. See
+// nearestRankIndex.
 func (rs *ReportService) calculateMetricsAggregates(metrics []types.ProcessedMetric) map[string]types.MetricAggregate {
 	metricsByLabel := make(map[string][]float64)
 
--- a/internal/services/report/report_service.go
+++ b/internal/services/report/report_service.go
@@ -2,6 +2,8 @@ package report
 
 import (
 	"fmt"
+	"math"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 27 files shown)


internal/services/report/report_service.go
```diff
@@ -645,48 +647,73 @@ func (rs *ReportService) flattenMetrics(cluster types.DiscoveredCluster) types.P
 	}
 }
 
-// calculateMetricsAggregates calculates avg, max, min aggregates for each metric type
+// CalculateMetricsAggregates is the exported entry point for callers outside
+// the report service (e.g. PlanService) that need percentile/aggregate
+// computations on flattened ProcessedMetric data.
+func CalculateMetricsAggregates(metrics []types.ProcessedMetric) map[string]types.MetricAggregate {
```
ifeeneyconfluent: Possible simplification: CalculateMetricsAggregates wrapper

  One thing worth considering here — since ReportService is an empty struct (type ReportService struct{}, line 20) and calculateMetricsAggregates doesn't reference rs, the receiver is effectively vestigial. That means this exported wrapper is mostly a way to spell NewReportService().calculateMetricsAggregates(...) for external callers, allocating a throwaway ReportService each call.

  Would it be worth promoting the function to package-level and dropping the wrapper? Something like:

  func CalculateMetricsAggregates(metrics []types.ProcessedMetric) map[string]types.MetricAggregate {
      // ...existing body, no receiver...
  }

  The internal call sites (report_service.go:266, 319, 370) and tests (report_service_test.go:884, 902, 917) would shift from rs.calculateMetricsAggregates(...) to CalculateMetricsAggregates(...), and plan_service.go:63 would stay as-is since it already uses the exported name.

  Functionally identical, just one fewer layer — happy to leave it as-is if you'd rather keep the method form for consistency with the other rs.* helpers.

ifeeneyconfluent (changes requested):

internal/services/report/report_service.go
suhothayan: Thanks for the feedback, fixed and pushed an update.

ifeeneyconfluent (approved):
LGTM

PR #295, opened 2026-05-15
feat: add `kcp scan connect-topics` for Kafka Connect worker discovery

adrian-januzi:
## Summary

- New `kcp scan connect clusters` subcommand that reads one or more topics playing the role of Kafka Connect's `connect-status` storage topic on a Kafka cluster, extracts `worker_id` values from `status-connector-*` / `status-task-*` messages, and prints the unique Connect worker addresses to stdout. Targets a single cluster via `--cluster-id` (auto-detects `--source-type` from the ID shape, matching `scan self-managed-connectors`) and reuses the existing `osk-credentials.yaml` / `msk-credentials.yaml` files for auth via the existing `sources.Source` abstraction.
- `--state-file` is required and validated to load cleanly but is **not** mutated in this phase — the flag is reserved for a later phase that will persist discovered addresses.
- New `internal/services/connectdiscovery/` package implements a narrow compacted-topic reader using sarama's consumer; new `Source.GetKafkaAdminForCluster` and `KafkaAdmin.BrokerConfig` accessors expose a single-cluster admin + its sarama config without growing the existing `Scan` path.

## Why

Operators planning a Confluent Cloud migration know which Kafka clusters they need to move but not always which Kafka Connect clusters run against them. `kcp scan self-managed-connectors` only inventories a Connect cluster the operator already has a REST URL for — useful for known clusters, useless for discovery. Parsing the `worker_id` field from the source Kafka cluster's `connect-status` topic surfaces the set of Connect workers (and therefore Connect clusters) talking to that Kafka cluster, with no prior knowledge of their REST endpoints.

This is the narrower successor to the topic-parsing connector discovery removed in #289. That path lived implicitly inside `kcp scan clusters`, reconstructed full connector configs and statuses, conflated cluster scanning with connector inventory, and required `kafka-cluster:ReadData` on the IAM policy of `scan clusters` itself. The new command is explicit, opt-in, scoped to its own IAM annotation (so `scan clusters` IAM stays trim), and only extracts `worker_id`s — no per-connector status reconstruction, no timestamp dedup, just the set of addresses.

A few UX details worth flagging:

- **Tombstones.** Compacted `connect-status` topics carry empty-value messages as deletion markers for connectors and tasks that no longer exist. The reader distinguishes these from real parse failures via a separate `Tombstones` counter and stays silent at log level — operators don't see alarming "JSON parse failed" lines for what's just deletion history.
- **Progress feedback.** Sarama's `NewConsumer` + metadata fetch + `OffsetOldest` seek can take 10–20 seconds before any messages flow. INFO logs at "connecting to brokers", "connected; beginning topic scan", "reading topic", and a per-partition completion line ensure the operator sees steady forward motion rather than wondering whether the CLI has hung.
- **No scheme inference.** Output is the raw `worker_id` (typically `host:port`) without prepending `http://`. Connect workers occasionally advertise schemes themselves; operators pick the scheme when piping into `scan self-managed-connectors`.

## Test plan

- [ ] `make test-go` passes (verified locally — `./cmd/scan/connect/...`, `./internal/services/connectdiscovery/...`, `./internal/sources/...`, `./internal/client/...` all green).
- [ ] `make test-kafka-connect` passes. **This is the primary integration-level regression net** — `integration-tests/osk-scan/run-connect.sh` now runs `kcp scan connect clusters --topics connect-status` against the docker-compose Connect profile after the existing `scan self-managed-connectors` test, asserts the worker address `localhost:8083` appears in stdout, and verifies the state-file SHA-256 is unchanged across the run (locks in the phase-1 no-write contract). I was unable to run this locally — Docker daemon wasn't up — so it's the highest-confidence pre-merge gate.
- [ ] `make docs-gen` produces `docs/assets/command-reference/scan/connect/{index,clusters}.md` cleanly. `mkdocs.yml` uses `awesome-pages`, so no nav edits are needed; the new directory is picked up automatically.
- [ ] Operator UX sanity check: run `kcp scan connect clusters --source-type osk --credentials-file osk-credentials.yaml --state-file kcp-state.json --cluster-id <some-id> --topics connect-status --verbose` against a real OSK cluster with a Connect deployment and confirm (a) the per-partition INFO logs appear, (b) the discovered worker addresses are emitted to stdout, (c) the state file's mtime is unchanged after the run.
- [ ] Auto-generated reference doc at `docs/assets/command-reference/scan/connect/clusters.md` reflects the command's Long description and IAM annotation.


integration-tests/osk-scan/teardown.sh
```diff
@@ -5,7 +5,11 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 ROOT_DIR="$(cd "$SCRIPT_DIR/../.." && pwd)"
 
 echo "Tearing down OSK scan broker..."
-docker compose -f "$SCRIPT_DIR/docker-compose.yml" down -v 2>/dev/null || true
+# --profile connect ensures profile-gated services (kafka-connect) are torn
+# down too. Without it, `docker compose down` only stops services with no
+# profile assigned, leaving the Connect container behind when run-connect.sh
+# fails partway through.
+docker compose -f "$SCRIPT_DIR/docker-compose.yml" --profile connect down -v 2>/dev/null || true
```
pinaruyar, line 12:
```suggestion
docker compose -f "$SCRIPT_DIR/docker-compose.yml" --profile * down -v 2>/dev/null || true
```

pinaruyar: so that we do not need to worry about any other profiles added in the future

internal/services/connectdiscovery/connectdiscovery.go
pinaruyar, line 1: What can be improved according to Claude. I am only putting it here in case you like performance coding and I kept seeing this closed-channel busy-loop bug. Feel free to ignore

🔴 1. Closed-channel busy-loop — line 174. The one real bug. case err := <-pc.Errors(): only returns on a non-nil error. When sarama closes that channel (which it does when the partition is done — confirmed in sarama's source), reading it returns nil forever, so the loop spins at ~100% CPU until the 30s timeout rescues it. Fix:
case err, ok := <-pc.Errors():
    if !ok {
        return // channel closed → partition done
    }
    slog.Warn("partition consumer error", "error", err)
    return

🟡 2. A fresh timer is allocated every loop iteration — line 181. case <-time.After(idleTimeout) builds a new timer on each pass (potentially per message). Hoist one timer and Reset it when a message arrives — that also removes the now-redundant re-check on line 182:
idle := time.NewTimer(idleTimeout)
defer idle.Stop()
// in the msg case: if !idle.Stop() { <-idle.C }; idle.Reset(idleTimeout)
// idle case becomes: case <-idle.C: return

🟢 3. Decode into a struct, not map[string]any — lines 213–220. You only read one field. A typed target is clearer, allocates less, and drops the .(string) assertion:
var payload struct {
    WorkerID string `json:"worker_id"`
}
if err := json.Unmarshal(msg.Value, &payload); err != nil { ... }
if payload.WorkerID == "" { stats.MissingField++; ...; return "", false }

🟡 4. Sequential drain — lines 92–95 and 123–146. Topics and partitions are read one at a time, and every partition pays the full 5s idle wait even when empty. A 5-partition connect-status ≈ 25s; multiply by topics. Draining partitions concurrently (an errgroup, writing into out under a sync.Mutex) would cut that to ~5s. The PR flags this as MVP-grade, so it's a reasonable defer — but worth a // TODO.

🟢 5. Timeouts are hard-coded constants — lines 37/42. Fine for now, but a slow or large cluster might need longer than 30s. Consider threading them through as options (the test seam already takes them as parameters; just expose them on ExtractWorkerIDs via an opts struct).

internal/sources/msk/msk_source.go
```diff
@@ -58,6 +58,60 @@ func (s *MSKSource) GetClusters() []sources.ClusterIdentifier {
...
+	}
+
+	clientBrokerEncryptionInTransit := utils.GetClientBrokerEncryptionInTransit(discoveredCluster.AWSClientInformation.MskClusterConfig)
+	kafkaVersion := utils.GetKafkaVersion(discoveredCluster.AWSClientInformation)
+
+	kafkaAdmin, err := createKafkaAdmin(authType, brokerAddresses, clientBrokerEncryptionInTransit, region, kafkaVersion, *clusterAuth)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Kafka admin: %w", err)
+	}
+	return *kafkaAdmin, nil
```
pinaruyar, line 112: heavily nitpicking here but this part is exactly same as scanCluster path [here](https://github.com/confluentinc/kcp/blob/f1c2748d8efb5665555ba7183b75416da2d2e950/internal/sources/msk/msk_source.go#L93). I would suggest to combine it in a helper function smth like **buildKafkaAdmin** but could be an overkill so I leave it to your discretion

cmd/scan/connect_topics/cmd_scan_connect_topics.go
```diff
@@ -0,0 +1,188 @@
...
+	topics = cleaned
+
+	// Validate the state file loads. Per R8 we do not mutate it. Skip when
+	// stateFile is empty so Cobra's MarkFlagRequired (which runs after PreRunE)
+	// can surface the canonical "required flag" error.
+	if stateFile != "" {
+		if _, err := os.Stat(stateFile); os.IsNotExist(err) {
+			return fmt.Errorf("state file does not exist: %s", stateFile)
+		}
+		if _, err := types.NewStateFromFile(stateFile); err != nil {
```
pinaruyar, line 140: duplicated below on line 147, should be okay to removed I don't think it would cause a time loss if it fails inside runScanConnectTopics

pinaruyar:
Screenshot 2026-05-31 at 20 37 12
looks like working properly

pinaruyar:
Q: Do we intend to mutate state file in the future or are we content with just printing the worker addresses?

adrian-januzi:
> Q: Do we intend to mutate state file in the future or are we content with just printing the worker addresses?

If I remember correctly, for this first pass it was purely just outputting the workers to the terminal and then we'd decide afterwards where it would go in the state file in a follow-up PR.

I think the rationale for deciding it later was that 1. we don't know how useful this will actually be; 2. we don't know how we'll use the follow-up `kcp scan connect-cluster` (or whatever we call it) command to read the connect clusters from the state file, if we even do that as the user will need to know the URL as that will also be the key in the state file. So they can't tell the command I want to use Connect cluster X without knowing the Connect URL for X.

pinaruyar (approved):

pinaruyar (approved):

PR #310, opened 2026-05-26
feat(lib): add pkg/lib bytes-in / bytes-out façade for external callers

suhothayan:
Adds `pkg/lib` — a public façade so external Go modules can call kcp's state-processing and plan-generation programmatically without reaching into `internal/*`.

```go
func ScanSummary(stateJSON []byte) ([]byte, error)
func GeneratePlan(stateJSON, planInputsYAML []byte) (*PlanResult, error)

type PlanResult struct {
    JSON       []byte // same as `kcp report plan --output json`
    Markdown   []byte // same as `kcp report plan --output md`
    PlanInputs []byte // resolved plan-inputs (request merged with kcp defaults), as YAML
}
```

- **Bytes in, bytes out.** `stateJSON` must be JSON (`kcp-state.json` shape). `planInputsYAML` must be YAML (`plan-inputs.yaml` shape) — pass `nil` to use defaults.
- **YAML echo.** `PlanInputs` comes back in `plan-inputs.yaml` shape so a UI can pre-populate an editable form from an empty initial call.
- **Stateless, concurrent-safe.** Each call constructs its own services.

## Calling from a service

```go
res, err := lib.GeneratePlan(stateBytes, planInputsYAML)
// res.JSON, res.Markdown, res.PlanInputs are all []byte — pass through to the client.
```

## Test plan

- [x] `go test ./pkg/lib/...` — covers happy path, malformed input rejection, default-echo on nil input, YAML plan-inputs parsing, and `plan.inputs` ↔ `PlanInputs` consistency (including a regression guard that the resolved `Raw` wrapper never appears in the YAML echo).

Marked **experimental** in `doc.go`; pin a specific kcp version in your `go.mod` until `plan_schema_version` leaves `"1-experimental"`.

### DIFF
--- a/cmd/create_asset/target_infra/testdata/iam_annotation.golden.md
+++ b/cmd/create_asset/target_infra/testdata/iam_annotation.golden.md
@@ -29,6 +29,7 @@
         "ec2:DescribeVpcs",
         "ec2:RevokeSecurityGroupEgress",
         "route53:AssociateVPCWithHostedZone",
+        "route53:ChangeResourceRecordSets",
         "route53:ChangeTagsForResource",
         "route53:CreateHostedZone",
         "route53:DeleteHostedZone",
--- a/internal/services/plan/render_markdown.go
+++ b/internal/services/plan/render_markdown.go
@@ -20,7 +20,14 @@ func RenderMarkdown(p *types.Plan, cfg *PlanConfig) ([]byte, error) {
 	var b bytes.Buffer
 
 	fmt.Fprintf(&b, "# Migration Plan — %s → Confluent Cloud\n\n", p.Header.Source)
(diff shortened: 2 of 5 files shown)


adrian-januzi (approved):

adrian-januzi:
Looks good though I am wondering what should be our process if there are breaking changes and how best to report them? My only worry is if we were to make some drastic changes (however unlikely), we might break the downstream services importing this? I guess this will be managed by the pinning of a specific package version, right?

suhothayan:
Yes, I assumed kcp will always support all previous kcp-state.json files, and in case if we are making a breaking change, kcp should fail loudly rather than silently misinterpret an older file: return an error like `kcp-state.json is incompatible with this version of kcp, regenerate the state file with kcp ≥ vX.Y.Z`, naming both the incompatibility and the minimum kcp version that produces a compatible file.

adrian-januzi:
Yup. We typically try to support backwards compatibility if there are no breaking changes and fail loudly if there is. I have also started looking into ensuring the full error message would be propagated down external services importing say this package to make sure this reaches downstream.

adrian-januzi (approved):

PR #317, opened 2026-06-01
feat(plan): red flags, effort signals, tiered storage, cost & cutover

suhothayan:
## Summary

Adds five new sections to `kcp report plan` plus per-cluster cutover overrides.

## New sections

- **§Red Flags** — 15 boolean trigger rows over scan + plan inputs (Kafka-version floor, IAM, Glue SR, Connect topics, MSK Express tier, tiered storage, EOS / Streams declarations, multi-region source, etc.). Triggered rows render at the top; NotTriggered/Unknown collapse into a tail summary.
- **§Effort Signals** — counters for Connect connectors, MM2 source aliases, self-managed Connect fleets, Glue serializer clients. `*int` distinguishes "unobservable" (nil) from "scan ran zero" (0).
- **§Tiered Storage** — surfaces clusters with `StorageMode == TIERED` and the customer's `consumer_history_requirement` / `historical_data_strategy` cascade. Reads CloudWatch Maximum (not Average) for RemoteLogSizeBytes.
- **§Cost vs Inventory Reconciliation** — parses AWS Cost Explorer usage strings (`<REGION>-Kafka.<family>.<size>` / `<REGION>-Express.<family>.<size>`) and lists clusters in the cost report but not in discovery.

## Per-cluster cutover overrides

- New inputs: `clusters[<name>].downtime_tolerance` / `.sub_pattern`. Heterogeneous fleets no longer need to be sliced and re-run.
- New `Plan.CutoverOverrides []ClusterCutoverOverride` — only diverging clusters appear. JSON carries structural `override_rejected` + `rejected_override_value` when a value is unrecognised; renderer adds a `*` marker.
- Typo'd values fire per-cluster OQs; overrides against an unknown cluster name fire `cluster_override_unknown_cluster` instead of being silently dropped.
- Per-cluster `seconds_per_service` against a non-mediated fleet fires `downtime_tolerance_requires_gateway` — gateway prereqs are fleet-scoped, so the per-cluster intent can't earn its own gateway.

## Other

Maintainability improvements across the plan package (no behaviour change).

## JSON contract

`PlanSchemaVersion` stays at `"1"`. All new fields are additive top-level keys (`cutover_overrides`, `override_rejected`, `rejected_override_value`) — no renames or removals.

## Test plan

- [x] `go test ./...` green
- [x] `golangci-lint run ./...` clean
- [x] 40 generated plans across 4 fleet shapes × 10 plan-inputs scenarios; per-cluster overrides, typo OQs, and rejected-override paths verified

### DIFF
--- a/internal/services/plan/red_flags.go
+++ b/internal/services/plan/red_flags.go
@@ -1,0 +1,600 @@
+package plan
+
+import (
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/confluentinc/kcp/internal/types"
+
+	kafkatypes "github.com/aws/aws-sdk-go-v2/service/kafka/types"
+)
+
+// Stable Red Flag IDs. Matches the spec's row numbering — kept stable
+// so JSON consumers can branch on the ID without parsing the title.
+const (
+	RedFlagIDSchemalessSource          = "schemaless_source"
+	RedFlagIDKafkaVersionBelowCLFloor  = "kafka_version_below_cl_floor"
+	RedFlagIDIAMAuthEnabled            = "iam_auth_enabled"
+	RedFlagIDGlueSRInUse               = "glue_sr_in_use"
+	RedFlagIDPartitionApproachingCap   = "partition_approaching_cap"
+	RedFlagIDMSKConnectPresent         = "msk_connect_present"
+	RedFlagIDSelfManagedConnectPresent = "self_managed_connect_present"
+	RedFlagIDMultiRegionSource         = "multi_region_source"
+	RedFlagIDZeroACLsWithIAM           = "zero_acls_with_iam"
+	RedFlagIDClientInventoryGap        = "client_inventory_gap"
+	RedFlagIDMSKExpressBrokerTier      = "msk_express_broker_tier"
+	RedFlagIDTieredStorageInUse        = "tiered_storage_in_use"
+	RedFlagIDEOSInUse                  = "eos_in_use"
+	RedFlagIDKafkaStreamsInUse         = "kafka_streams_in_use"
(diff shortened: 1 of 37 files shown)


internal/services/plan/red_flags.go
```diff
@@ -0,0 +1,600 @@
...
+	return rf
+}
+
+func brokerInstanceType(c types.ProcessedCluster) string {
+	prov := c.AWSClientInformation.MskClusterConfig.Provisioned
+	if prov == nil || prov.BrokerNodeGroupInfo == nil || prov.BrokerNodeGroupInfo.InstanceType == nil {
+		return ""
+	}
+	return *prov.BrokerNodeGroupInfo.InstanceType
+}
```
adrian-januzi, line 469: Just thought I'd flag this - Serverless clusters will never have `Provisioned` not equal to `nil`. Instead Serverless clusters populate their cluster's details inside a `Serverless` object:

```json
              "Provisioned": null,
              "Serverless": {
                "VpcConfigs": [
                  {
                    "SubnetIds": [
                      "subnet-xxxxx",
                      "subnet-xxxxx"
                    ],
                    "SecurityGroupIds": ["sg-xxxxx"]
                  }
                ],
                "ClientAuthentication": {
                  "Sasl": { "Iam": { "Enabled": true } }
                }
              },
```

Just conscious as I saw mention of Serverless clusters elsewhere in this PR.

adrian-januzi (approved):
lgtm to me

internal/services/plan/red_flags.go
suhothayan: Thanks for pointing it out, I'll handle this in a follow up PR.

PR #318, opened 2026-06-01
fix: remove duplicate error output in CLI

bmartin494:
## Summary

- When a command failed, the error was printed twice: once via `slog.Error()` (which writes to both the log file and console via FanOutHandler) and once via `fmt.Fprintf(os.Stderr, ...)`. Removed the redundant `fmt.Fprintf` line from `main.go`.
- Extracted `run()` from `main()` to make error handling testable without `os.Exit`.
- Added `TestCommandErrorNotDuplicated` which triggers a command error, captures stdout+stderr, and asserts the error appears exactly once. The test uses the returned error text dynamically so it won't break if error wording changes.

## Test plan

- [x] `TestCommandErrorNotDuplicated` fails before fix (error count = 2), passes after (count = 1)
- [x] Full test suite passes (`make test-go`)
- [x] Manual verification: `kcp scan clusters --source-type invalid --credentials-file foo.yaml` shows error once

### DIFF
--- a/cmd/cmd_root.go
+++ b/cmd/cmd_root.go
@@ -27,9 +27,10 @@ import (
 var verbose bool
 
 var RootCmd = &cobra.Command{
-	Use:   "kcp",
-	Short: "A CLI tool for kafka cluster planning and migration",
-	Long:  "A comprehensive CLI tool for planning and executing kafka cluster migrations to confluent cloud. Docs: " + build_info.DocsURL(),
+	Use:           "kcp",
+	Short:         "A CLI tool for kafka cluster planning and migration",
+	Long:          "A comprehensive CLI tool for planning and executing kafka cluster migrations to confluent cloud. Docs: " + build_info.DocsURL(),
+	SilenceErrors: true,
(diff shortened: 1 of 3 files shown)


adrian-januzi:
I think we want to set `SilenceErrors: true` in `cmd/cmd_root.go` so that we don't get double-errors at the root level too:

```shell
❯ ./kcp test
Error: unknown command "test" for "kcp"
Run 'kcp --help' for usage.
2026/06/02 11:46:20 ERROR unknown command "test" for "kcp"
```

bmartin494:
> I think we want to set `SilenceErrors: true` in `cmd/cmd_root.go` so that we don't get double-errors at the root level too:
> 
> ```shell
> ❯ ./kcp test
> Error: unknown command "test" for "kcp"
> Run 'kcp --help' for usage.
> 2026/06/02 11:46:20 ERROR unknown command "test" for "kcp"
> ```

@adrian-januzi yeah good catch thanks, have pushed the change for this now

adrian-januzi (approved):
lgtm

adrian-januzi (approved):

PR #320, opened 2026-06-03
feat: gov edition (kcp-lite) via Go build tags

adrian-januzi:
## Summary

Adds a `gov` Go build tag that compiles a slimmed **`kcp-lite`** binary, stripping the three `create-asset` commands a gov target environment can't honor — `target-infra`, `migration-infra`, and `migrate-connectors` — via a self-registration registry with tag-gated imports. A tag-derived `build_info.Mode` constant surfaces the edition in CLI output and as a GOV banner in the web UI. The default build is unchanged (`prod`, full `kcp`).

The driver is **product fit, not an audit boundary** — there's no claim the stripped code is provably absent for a binary scan.

> **Release counterpart:** the GoReleaser dual-edition builds live in confluentinc/kcp-release#57, which produces the `kcp-lite_{os}_{arch}` artifacts this branch's self-update path resolves.

## What changed

| Area | Change |
|---|---|
| `internal/build_info` | `Mode` const per tag-gated file (`mode_prod.go` `!gov`, `mode_gov.go` `gov`) + `IsGov()`. Derived from the tag, never an ldflag, so the reported edition can't disagree with what's compiled in. |
| `cmd/create_asset` | Static `AddCommand` list replaced by a self-registration registry. The 3 gov-excluded packages are imported only from `register_full.go` (`//go:build !gov`), so under `-tags=gov` they're unreferenced and **absent from the binary** — not merely unregistered. |
| `cmd/version`, `cmd/cmd_root` | Edition surfaced in `kcp version` and the startup banner. |
| `cmd/ui` | `GET /edition` endpoint + `GovBanner` React component (fails safe to hidden on error / non-gov). |
| `cmd/update` | Self-update resolves the release asset by edition (`kcp` vs `kcp-lite`) and refuses rather than installing the wrong edition if no match. |
| `Makefile` | `build-gov` (→ `kcp-lite`) and `verify-gov` (compile + vet + test under `-tags=gov`). |
| `.semaphore` | `verify-gov` wired into CI so tag-gated breakage is caught. |

## Testing

- Full prod `go test ./...`, `go vet ./...`, and `golangci-lint` — all green.
- `make verify-gov` — compiles, vets, and runs gov-tagged tests under `-tags=gov`.
- **Stripping proven**: `go tool nm` / `strings` on the gov binary show **0** references to the three packages vs **170** in prod; `kcp-lite create-asset --help` omits them.
- Self-update filter: table tests cover prod/gov × linux/darwin/windows, archive and cross-edition rejection, and the download-URL form. The prod filter was verified against the live release assets.
- UI: Go endpoint test (both tags) + Playwright covering banner-hidden (prod) and banner-shown (stubbed `/edition=gov`).

## Known Residuals

- **[P2 / testing]** No integration test for `Updater.Run()`'s edition-aware `!found` refusal path. The pure `assetFilterFor` is fully table-tested and the path fails safe; covering `Run()` directly needs a refactor to inject a mock `go-selfupdate` Source. Accepted as a follow-up — core asset-resolution logic is tested.

## Post-Deploy Monitoring & Validation

- **After the next release** (with [kcp-release#57 merged](https://github.com/confluentinc/kcp-release/pull/57)): confirm a `kcp-lite` binary reports `Edition: gov`, omits the three commands, and `kcp update` from a `kcp-lite` binary resolves the `kcp-lite_*` asset (not the full `kcp`). Confirm prod `kcp update` still resolves `kcp_*`.
- **Failure signal:** a `kcp-lite` self-update that installs `kcp`, or refuses when a matching asset exists; the GOV banner appearing on a prod build.
- **Rollback:** the `gov` tag is opt-in and the default build is byte-unaffected, so reverting touches no prod behavior.
- **Owner / window:** release owner, first tagged release after both PRs merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ migration-plan/
 # Ignore kcp generated files
 kcp-scan/
 kcp
+kcp-lite
 set_migration_env_vars.sh
 kcp-scan*
 cluster-credentials.yaml
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -74,6 +74,10 @@ blocks:
             - cd cmd/ui/frontend && yarn install && yarn build && cd -
             - (cd cmd/ui/frontend && npx playwright install chromium chromium-headless-shell --with-deps)
             - make test-go GOTEST_FLAGS="-p 1 -parallel 4 -count=1"
+            # Guard the gov build tag: compile, vet, and test the -tags=gov
+            # variant so tag-gated breakage (only the prod tree builds by
+            # default) is caught here.
+            - make verify-gov GOTEST_FLAGS="-p 1 -parallel 4 -count=1"
             - make test-playwright
       epilogue:
         on_pass:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,9 +56,21 @@ make build-linux
 make build-darwin-arm64
 make build-all
 
+# Build the slimmed "gov" edition as kcp-lite (see Editions below)
+make build-gov
+
 # Install to system path (requires sudo)
 make install
 ```
+
+### Editions
+
+KCP compiles into two editions selected by the `gov` Go build tag:
+
+- **prod** (default, `make build` → `kcp`) — the full command tree.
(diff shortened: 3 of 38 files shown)


adrian-januzi:
Closed in favour of https://github.com/confluentinc/kcp/pull/330

PR #327, opened 2026-06-08
fix: generate providers.tf for migrate-connectors and add SASL mechanism to migration execute

bmartin494:
## Summary

Fixes two bugs reported against v0.8.1:

- **`create-asset migrate-connectors` missing `providers.tf`** — The MSK and self-managed connector migrators rendered per-connector `.tf` files but never wrote `providers.tf` or `variables.tf`, causing `terraform apply` to fail with "Could not retrieve the list of available versions for provider hashicorp/confluent". Every other `create-asset` subcommand already generated these files via the HCL service.
- **`migration execute --use-sasl-scram` defaults to SHA-256** — The SASL/SCRAM mechanism field was never set, so the empty string defaulted to SHA-256 in the Kafka client. MSK only supports SHA-512, causing "ran out of available brokers" errors. This was a regression introduced when migration commands were decoupled from AWS MSK.

### Changes

**Commit 1 — providers.tf fix:**
- Export `GenerateProvidersTf()` and `GenerateMigrateConnectorsVariablesTf()` on `MigrationScriptsHCLService`
- Fill in the previously-empty variables stub with Confluent Cloud API key/secret variables
- Add `WriteMigrateConnectorsInfraFiles()` shared helper to eliminate duplication between MSK and self-managed migrators
- Remove dead code (`generateMigrateConnectorsFiles`, `generateMigrateConnectorsMainTf`) that had no callers

**Commit 2 — SASL mechanism fix:**
- Add `--sasl-scram-mechanism` flag (default `SHA512`) to `migration execute`
- Validate the flag value early in `PreRunE` (must be `SHA256` or `SHA512`)
- Thread the mechanism through `MigrationExecutorOpts` into `SASLScramConfig.Mechanism`

**Commit 3 — SCRAM-SHA-512 integration test variant:**
- Enable `SCRAM-SHA-512` on the existing OSK scan broker alongside SHA-256
- Register SHA-512 credentials for the `admin` and `kafkauser` accounts
- Add `kafka-sasl-sha512.yaml` credential file and include `sasl-sha512` in the `run.sh` test loop

**Commit 4 — dedicated SCRAM-SHA-512-only broker:**
- Add a standalone Kafka broker (port 9099) with only `SCRAM-SHA-512` enabled — no SHA-256 fallback
- Includes producer/consumer containers to match the other standalone brokers
- If a client accidentally defaults to SHA-256, this broker rejects it outright, reproducing the exact failure from the bug report

**Commit 5 — terraform validation for connectors and Glue schemas:**
- Add `terraform init` + `terraform validate` tests for migrate-connectors and migrate-glue-schemas output
- Closes the testing gap that allowed the missing `providers.tf` bug to ship
- Fixes a subtle issue where the Glue schemas test passed vacuously against an empty directory (the `schemaProjectToFiles` helper didn't include `AdditionalFiles`)

### Follow-ups

- **Migration e2e SASL/SCRAM scenario** — The migration e2e tests (`integration-tests/migration/`) only exercise `--use-unauthenticated-plaintext`. Adding a SASL/SCRAM-SHA-512 scenario would require provisioning a new source broker with SCRAM auth, cluster link configuration, and gateway CR fixtures. Tracked as a follow-up.

## Test plan

- [x] `make test-go` — all existing + new tests pass, zero failures
- [x] New unit tests for providers.tf/variables.tf generation in both MSK and self-managed connector migrators
- [x] New tests for SASL mechanism: default SHA512, explicit SHA256 override, env var binding, invalid value rejection, end-to-end slog capture
- [x] Manual verification: ran `create-asset migrate-connectors msk` on main (empty output dir) vs fix branch (providers.tf + variables.tf generated with correct Confluent provider declaration)
- [x] `make test-osk-scan` — all 13 Docker integration tests pass including both SHA-512 variants (shared broker + SHA-512-only broker), all reporting 5 topics, 12 ACLs
- [x] `terraform validate` tests for connectors and Glue schemas — both pass `terraform init` + `terraform validate` with the Confluent provider


cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -109,6 +110,7 @@ the migration state file and must be provided each time.`,
 	saslScramFlags.SortFlags = false
 	saslScramFlags.StringVar(&saslScramUsername, "sasl-scram-username", "", "SASL/SCRAM username for the source MSK cluster.")
 	saslScramFlags.StringVar(&saslScramPassword, "sasl-scram-password", "", "SASL/SCRAM password for the source MSK cluster.")
+	saslScramFlags.StringVar(&saslScramMechanism, "sasl-scram-mechanism", "SHA512", "SASL/SCRAM mechanism (SHA256 or SHA512). Defaults to SHA512 for MSK compatibility.")
```
adrian-januzi, line 113: Thoughts on removing the default and forcing the user to think about their SHA mechanism?

I'm 50/50 on the idea so I'll let you decide if you wanna apply that. I can see the value in also just keeping it with a default.

adrian-januzi (approved):

adrian-januzi (approved):

bmartin494:
Landed via squashed commit 8cd8f07 on main, included in v0.8.2 and v0.8.3. Closing as the code was pushed directly rather than merged through the PR.

PR #326, opened 2026-06-08
refactor(types): move migration FSM state into internal/services/migration (Pass 2.2)

ifeeneyconfluent:
## Summary

PR 2 of Pass 2 (moving owner-specific types out of `internal/types` into their consumer packages).

Relocates the **migration FSM state file types** into their sole owning service, `internal/services/migration`:

- **Moved** → `internal/services/migration/state.go`: `MigrationState`, `MigrationConfig`, `NewMigrationState`, `NewMigrationStateFromFile`, `WriteToFile`/`UpsertMigration`/`GetMigrationById`, and the FSM `State*`/`Event*` constants. The state round-trip test moved alongside.
- **Stays** in `internal/types`: `MigrationType` + `Manifest`. These are *infra-type config* shared by `hcl`/`create-asset`, **not** FSM execution state — moving them into the migration service would risk an `hcl → migration` import cycle (flagged during Pass 2 validation).

Importers updated: `cmd/migration/{execute,init,list}` now reference `migration.MigrationState` etc.; the migration service's own files drop the `types.` qualifier. `MigrationState.KcpBuildInfo` stays `types.KcpBuildInfo` (shared), so the service keeps a one-symbol dependency on `internal/types` — no cycle (`internal/types` does not import the service).

## Test Plan

- [x] `go build ./...`
- [x] `go test ./...` — 47 packages pass
- [x] `golangci-lint run --config .golangci.yml ./...` (v2.11.3, `new-from-rev: main`) — 0 issues
- [x] `gofmt -l` clean

## Pass 2 sequence

PR 2 of ~7. Done: gateway (#325). Next: schema_registry → `schema_registry`, then terraform → `hcl`, wizard_requests → `cmd/ui/api`, plan → `plan`, processed (partial) → `report`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #325, opened 2026-06-08
refactor(types): remove dead gateway CR structs (Pass 2.1)

ifeeneyconfluent:
## Summary

First step of **Pass 2** of the `internal/types` reorg (moving owner-specific types out into their consumer packages).

While validating the planned `gateway.go → internal/services/gateway` move, I found the structs are **dead code**: `GatewayResource`/`GatewaySpec`/`Route`/etc. were added in #224 but `types.GatewayResource` has **never been referenced** anywhere in the git history. The real gateway custom-resource handling works with YAML templates + k8s unstructured objects (`unstructured.NestedSlice(spec, "streamingDomains")`), not these Go structs.

So rather than relocate dead scaffolding into the gateway service, this deletes it — which advances the Pass 2 goal (slim `internal/types`) more directly.

## Test Plan

- [x] `go build ./...` — unaffected
- [x] `go vet ./internal/types/` clean
- [x] `go test ./...` — unaffected
- [x] `git log -S 'types.GatewayResource' --all` — zero hits (confirms never used)

## Pass 2 sequence

This is PR 1 of ~7 sequential per-domain PRs. Remaining: migration FSM → `migration`, schema_registry → `schema_registry`, terraform → `hcl`, wizard_requests → `cmd/ui/api`, plan → `plan`, processed (partial — `ProcessedClusterMetrics` stays due to a state-schema cycle) → `report`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #328, opened 2026-06-08
refactor(types): move Terraform models to internal/services/hcl/hcltypes (Pass 2.3)

ifeeneyconfluent:
## Summary

PR 3 of Pass 2. Relocates the **Terraform/HCL data models** out of `internal/types`.

The plan's destination was the parent `internal/services/hcl` package, but that **causes an import cycle**: the parent `hcl` package imports its subpackages (`modules`, `confluent`, `aws`), and those subpackages are the heaviest users of these types — so `hcl/modules → hcl` (for the types) plus the existing `hcl → hcl/modules` is a cycle.

Instead, the types go into a new **dependency-free leaf package** `internal/services/hcl/hcltypes`, which the parent `hcl`, all subpackages, and external callers (`cmd/create_asset`, `cmd/ui/api`, `internal/utils`) import without any cycle.

Moved: `TerraformState`, `TerraformOutputOld`, `TerraformOutputValue`, `TerraformFiles`, `TerraformVariable`, `TerraformOutput`, `MigrationInfraTerraform*`, `MigrationScriptsTerraform*`. ~169 references rewritten `types.X` → `hcltypes.X` (imports fixed with goimports).

No back-edge: `internal/types` referenced none of these (they aren't part of the persisted `State`), so they leave the shared package cleanly.

One pre-existing unused method (`generateMigrateConnectorsFiles` — WIP, not wired into the orchestrator) got a `//nolint:unused`; the `new-from-rev` linter only surfaced it because the type rename touched its signature line. Flagged for separate cleanup.

## Test Plan

- [x] `go build ./...`
- [x] `go test ./...` — 47 packages pass
- [x] `golangci-lint run --config .golangci.yml ./...` (v2.11.3, `new-from-rev: main`) — 0 issues
- [x] `gofmt -l` clean

## Pass 2 sequence

PR 3 of remaining. Done: gateway (#325 ✅), migration (#326 ✅). **Skipped** schema_registry (validation: it's shared state-schema, stays in `internal/types`). Next: wizard_requests → `cmd/ui/api`, plan → `plan`, processed (partial) → `report`.

### DIFF
--- a/cmd/create_asset/bastion_host/bastion_host_generator.go
+++ b/cmd/create_asset/bastion_host/bastion_host_generator.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/confluentinc/kcp/internal/services/hcl"
+	"github.com/confluentinc/kcp/internal/services/hcl/hcltypes"
 	"github.com/confluentinc/kcp/internal/types"
 	"github.com/confluentinc/kcp/internal/utils"
 )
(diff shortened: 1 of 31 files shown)


ifeeneyconfluent:
Superseded by a fresh re-execution off latest `main` (this branch no longer rebased cleanly after the Go version bump, connector-scanning removal, and CC-gov gating landed). Same `hcltypes` leaf-subpackage approach. See the new PR.

PR #334, opened 2026-06-10
feat(discover): add --cluster-arn flag for single-cluster discovery

ifeeneyconfluent:
## Summary
- Adds `--cluster-arn` to `kcp discover` to (re)discover specific MSK cluster(s) without overwriting the other clusters in their region — enabling per-cluster metrics granularity.
- `--region` and `--cluster-arn` are mutually exclusive (exactly one required); in targeted mode the region(s) are inferred from the ARNs. Region-level discovery (list clusters, costs, configs) still runs; only the per-cluster loop is filtered to the targeted ARNs.
- New `UpsertTargetedClusters` methods (state + credentials) create-or-replace only the targeted clusters and preserve siblings, merging scan-acquired data (ACLs/auth) via the existing merge logic. Full-region discovery behaviour is unchanged.
- Works from a fresh state file too: `kcp discover --cluster-arn <arn>` with no prior state yields a state file containing exactly that one cluster in its inferred region.

## Test plan
- `make test-go` green.
- Unit tests cover targeted upserts (create new region, replace existing cluster, append a new cluster into an existing region, sibling preservation, region cost/config refresh), the ARN region-derivation/filter helpers, and the single-vs-full persistence routing (`persistDiscoveredRegion`: full-region prunes removed clusters, targeted preserves siblings).
- Manual CLI checks: `--region`/`--cluster-arn` mutual-exclusivity, one-required, and malformed-ARN validation.

## Notes
- The discoverer's live-AWS orchestration is not unit-tested in this codebase (pre-existing); the new logic is factored into pure, tested helpers and type methods.
- Touches `internal/types`; may need a rebase against the in-flight types reorg.


adrian-januzi (approved):
LGTM - the `--metrics-granularity 60s` causes a bit of lag on the chart but that's out of scope (@tjunderhill nudge nudge).

PR #341, opened 2026-06-17
fix(security): write kcp state files with 0600 permissions

adrian-januzi:
## What & why

`kcp` wrote its state files — `kcp-state.json` and `.kcp-migration-state.json` — with mode **0644** (group/world-readable). These hold infrastructure-reconnaissance metadata: broker bootstrap endpoints, cluster topology, ACL principals/rules, topic and consumer-group names, metrics, and migration progress. On a shared host — notably the bastion hosts `kcp` itself provisions to reach private clusters — any other local account could read them.

The credentials YAMLs were already `0600`, so the state files were the odd ones out. This change writes both at **0600** (owner read/write only), closing unauthorized local reads of the infra map.

## How

Both writes funnel through two central choke points — `State.WriteToFile` (`internal/types/state.go`) and `MigrationState.WriteToFile` (`internal/services/migration/state.go`) — so every command that persists state (`discover`, `scan clusters` / `client-inventory` / `schema-registry`, `migration init` / `execute`) inherits the hardening with no call-site changes.

Each writer now:
1. creates the temp file via `os.CreateTemp` — a uniquely-named file created with `O_EXCL` at `0600`,
2. pins the mode with an explicit `Chmod(0600)`, then
3. atomically `os.Rename`s it onto the target.

Why not just flip the `0644` literal to `0600`? The old code wrote to a fixed-name `<path>.tmp`, and `os.WriteFile`'s mode only applies when it *creates* the file — so a stale temp from a crashed run would keep its old `0644` mode and the rename would carry that through. `os.CreateTemp` sidesteps this (fresh unique file at `0600`) and also eliminates the transient world-readable window a chmod-after-write would leave. The real state file is only ever swapped by the atomic rename and is never deleted directly, so a crash before the rename leaves the previous file intact.

## Impact / compatibility

- **Same user, no change:** `0600` keeps the owner's read/write; the atomic rename still works.
- **No migration needed:** an existing `0644` state file is tightened to `0600` on the next write.
- **Out of scope (intentional):** generated Terraform/HCL, report-command text, and docs stay `0644` — they are artifacts meant to be applied/shared, not secrets.

## Testing

Test-first (TDD). RED-verified the happy path (`0644 != 0600`) and the stale-temp abuse case (which forced the `os.CreateTemp` design). Coverage across `state_test.go`, `migration_state_test.go`, and the new `*_umask_unix_test.go` files:

- new file is `0600`
- a second write preserves `0600` (does not loosen back to `0644`)
- a stale `0644` temp does not leak into the result
- an existing `0644` file is tightened to `0600`
- umask cannot loosen the result below `0600`
- content still round-trips through the loader

Verification: full suite green (**46 ok, 0 fail**), `go vet` clean, `gofmt` clean, cross-platform build OK (darwin / linux / windows). On-disk behavior confirmed — a real `WriteToFile` produces `-rw-------`, and a second write keeps it `-rw-------`.

## Notes for reviewers

- The umask regression tests use `syscall.Umask` (unix-only) and live behind a `//go:build unix` constraint, so `go test` still compiles on Windows (kcp ships a `build-windows` target).
- Security review: no Critical/High/Medium findings — net improvement. One informational note: on a shared host where `--state-file` points at a world-writable directory, generic directory-level TOCTOU is a pre-existing environmental condition, neither introduced nor worsened here (posture strictly improves `0644` → `0600`).

### DIFF
--- a/internal/services/migration/migration_state_test.go
+++ b/internal/services/migration/migration_state_test.go
@@ -3,6 +3,7 @@ package migration
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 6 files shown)


adrian-januzi:
**Note on cross-machine workflows (bastion → laptop):** 0600 does not break the common "create the state file on a bastion host, then `scp` it to your laptop" flow, and needs no `chown`/`chmod` workaround.

`scp` copies file *contents*, not ownership — it can't stamp a foreign UID without root. So the local copy is owned by the laptop user running `scp`, and 0600 (`rw-------`) grants that owner full read/write. `kcp ui` and subsequent `kcp` commands run as that same user, so they just work. Plain `scp` also lands the copy at the laptop's umask (often 0644); the next kcp write re-creates it at 0600 anyway, so it self-heals.

The only case where 0600 differs from the old 0644 is when the copy ends up owned by a *different* user than the one running kcp (e.g. `sudo scp` landing it as root, or a genuinely shared multi-user host) — which is the intended protection, and atypical for this workflow.

bmartin494 (approved):
LGTM 🚢

PR #349, opened 2026-06-18
feat(healthcheck): add kcp healthcheck for live Kafka inventory (MSK + Apache Kafka, hidden)

bmartin494:
## Summary
- Adds a top-level `kcp healthcheck` command that scans a Kafka cluster via the Kafka Admin API and writes a markdown inventory report per cluster (cluster ID, brokers, topics, ACLs).
- Supports both **Apache Kafka** (`--source-type apache-kafka`, reads `apache-kafka-credentials.yaml`) and **MSK** (`--source-type msk`, reads `msk-credentials.yaml` + `kcp-state.json` populated by `kcp discover`). MSK SCRAM is pinned to SHA-512.
- Raw inventory only — no derived findings / severity. Tier 2 checks (ISR, consumer-group lag, etc.) are deferred.
- Command is **hidden from `--help` and `gen-docs`** while the migration-readiness framing and invocation model are still being decided. Wired up and callable so downstream struct refactors can rely on the surface; users won't discover it organically.
- Prints a per-cluster summary line + ✅ completion block to stdout so the operator sees confirmation in the terminal (same shape for OSK and MSK). Structured `slog.Info` record is also emitted to `kcp.log`.
- Re-uses the existing OSK/MSK sources, IAM-policy annotation helper, and markdown service rather than introducing a parallel scanning path.

## Test plan
- [x] `make test-go` — all packages green, including `cmd/healthcheck/` tests
- [x] End-to-end against `integration-tests/osk-scan` SASL/SCRAM Docker env → valid markdown report + stdout summary
- [x] End-to-end against MSK playground (`kcp discover` → `kcp healthcheck --source-type msk`) → valid markdown report for the discovered cluster
- [x] Missing state file for MSK → friendly error pointing to `kcp discover`
- [x] Command absent from `./kcp --help` output
- [x] Missing required flags produce Cobra's standard error (not a stack trace)
- [ ] Reviewer to spot-check the markdown layout

### DIFF
--- a/cmd/cmd_root.go
+++ b/cmd/cmd_root.go
@@ -12,6 +12,7 @@ import (
 	"github.com/confluentinc/kcp/cmd/create_asset"
 	"github.com/confluentinc/kcp/cmd/discover"
 	"github.com/confluentinc/kcp/cmd/docs"
+	"github.com/confluentinc/kcp/cmd/healthcheck"
 	"github.com/confluentinc/kcp/cmd/migration"
 	"github.com/confluentinc/kcp/cmd/report"
 	"github.com/confluentinc/kcp/cmd/scan"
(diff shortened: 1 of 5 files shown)


tjunderhill (approved):
LGTM!

PR #352, opened 2026-06-19
refactor(types): dedup MSK/OSK credentials auth-method + file-I/O

ifeeneyconfluent:
## Summary

Two small, independent, behavior-preserving commits that stop MSK and OSK from maintaining parallel copies of credential auth-method behavior and file-I/O boilerplate. Deferred final follow-up of the `internal/types` reorg. No YAML wire-format change.

The top-level `Credentials` (region/ARN-nested) vs `OSKCredentials` (flat + bootstrap + monitoring) shapes are **not** unified here — that is YAML-compat-sensitive and deferred to its own future plan.

## Commit 1 — dedup auth-method behavior onto `AuthMethodConfig`

MSK (`ClusterAuth`) and OSK (`OSKClusterAuth`) each had a near-identical `GetAuthMethods` and an exact-duplicate `GetSelectedAuthType`. Moved the canonical-order enumeration and single-method selection onto the **already-shared** `AuthMethodConfig`:

- `EnabledAuthMethods(includeIAM bool) []AuthType` — enabled methods in canonical order; `includeIAM=false` for sources without IAM (Apache Kafka / OSK).
- `SelectedAuthType(includeIAM bool) (AuthType, error)` — single enabled method or error.
- `ClusterAuth` delegates with `includeIAM=true`; `OSKClusterAuth` with `includeIAM=false` (preserves "IAM not supported for OSK").

IAM position (3rd) is preserved, so output is byte-identical to both prior implementations on every input.

## Commit 2 — extract credentials file-I/O boilerplate

`New*CredentialsFromFile` and the two `WriteToFile` methods duplicated the same read→unmarshal→`Validate` and marshal→`WriteFile(0600)` boilerplate. Extracted two shared helpers in new `credentials_io.go`:

- `writeYAMLFile(path, v)` — marshal + `os.WriteFile(…, 0600)`.
- `loadCredentialsFile[T interface{ Validate() (bool, []error) }](path)` — read + unmarshal + `Validate`. Both credential types satisfy the value-receiver constraint.

The four functions become one-line wrappers; public signatures unchanged. `Credentials.ToYaml()` retained.

**Intentional behavior change:** MSK's read error previously hardcoded `"failed to read creds.yaml file"` regardless of the actual path (a latent bug); the shared helper reports the real `path`. YAML output bytes and `0600` perms are unchanged.

## Verification

- `go test ./internal/types/` — ok (existing ~60 credentials tests + new focused tests pass unchanged)
- `go build ./...` — exit 0
- `go vet ./...` — exit 0
- `go test ./...` — **47 packages OK, 0 failures**

### DIFF
--- a/internal/types/auth_method_config_test.go
+++ b/internal/types/auth_method_config_test.go
@@ -1,0 +1,130 @@
+package types
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAuthMethodConfig_EnabledAuthMethods(t *testing.T) {
+	tests := []struct {
+		name       string
+		config     AuthMethodConfig
+		includeIAM bool
+		expected   []AuthType
+	}{
+		{
+			name:       "empty config returns no methods",
+			config:     AuthMethodConfig{},
+			includeIAM: true,
+			expected:   []AuthType{},
+		},
+		{
+			name: "single SASL/SCRAM method",
+			config: AuthMethodConfig{
+				SASLScram: &SASLScramConfig{Use: true},
+			},
+			includeIAM: true,
+			expected:   []AuthType{AuthTypeSASLSCRAM},
+		},
+		{
+			name: "IAM included when includeIAM is true",
+			config: AuthMethodConfig{
+				IAM: &IAMConfig{Use: true},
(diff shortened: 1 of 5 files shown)


adrian-januzi (approved):
lgtm

PR #354, opened 2026-06-22
feat(scan,discover): restore connector scanning with sensitive-config redaction

adrian-januzi:
## Summary

Re-introduces the connector discovery and state persistence removed in the 0.8.3 release (commit `5b93270b`), hardened with an always-on redaction layer so connector secrets never reach `kcp-state.json` or `kcp.log`. The `migrate-connectors` consumer — whose types and read paths survived the removal — works against the re-populated state unchanged.

## What changed

**New `internal/redact/` package.** A single, always-on, fail-closed redaction utility used by both discovery paths:
- Case-insensitive **substring** blacklist: ~120 sensitive config names (derived from `confluentinc/connect-migration-utility` `config_discovery.py`) plus the patterns for password / token / secret / credential.
- Recurses into nested maps/lists; replaces sensitive values with the literal `<kcp-redacted>` (keys preserved).
- No opt-out, no user extension, one-way.

**`kcp discover` — MSK Connect discovery restored.** `ListConnectors` / `DescribeConnector`, matched to discovered clusters by bootstrap servers, persisted as `ConnectorSummary`. The `kafkaconnect:ListConnectors` / `kafkaconnect:DescribeConnector` IAM permissions are re-added to the discover IAM annotation (golden updated).

**`kcp scan self-managed-connectors` restored.** Standalone command hitting the Kafka Connect REST API with SASL/SCRAM, mTLS, or unauthenticated access; persists `SelfManagedConnectors`.

**Redaction seam.** Both paths redact connector config **before** it enters any persisted struct *and* **before** any logging — logs report counts only (e.g. "redacted N fields across M connectors"), never key names or values.

## Testing & verification

- `go test ./...` — **50 packages ok, 0 failures**
- `go vet ./...` — clean · `go build ./...` — clean · `gofmt` — clean

New test coverage (TDD, RED-first):
- Redaction matching: exact / substring / prefixed / case-insensitive / nested; **no false positives** on benign keys (`tasks.max`, `connector.class`, `key.converter`); empty maps; exact redaction counts.
- MSK Connect: `AccessDenied` on `ListConnectors` **warns and continues** (non-fatal); non-matching connectors excluded; **incomplete/partial connector skipped** (no panic); redact-before-persist + redact-before-log.
- Self-managed scan: basic-auth vs unauthenticated header selection (httptest); non-200 → error; flag mutual-exclusivity + required-auth validation; bad TLS cert path → clean error (no panic).
- End-to-end: no raw secret survives `PersistStateFile` on either shape; `migrate-connectors` access paths read the re-populated (redacted) state after a round-trip.

## Robustness fixes (from review)

Two latent panics were closed during review, both via failing-test-first:
- A partial AWS connector response (nil required fields) is now **skipped with a warning** rather than panicking and aborting `discover`.
- A bad TLS cert path now returns a **clear error** from the scanner constructor instead of a later nil-pointer panic.

## Follow-ups (warning, pagination, merge)

A second pass closes three gaps that undermined the redaction work, built test-first (RED→GREEN per behavior) and run through dual code + security review:

- **Redacted-config warnings.** `kcp create-asset migrate-connectors` (both MSK and self-managed) now prints a **count-based** warning when generated assets contain `<kcp-redacted>` placeholders the operator must replace before `terraform apply` — never connector names or field keys. A matching banner appears in the UI connector explore view (both tabs). Backed by reusable `redact.MapContainsRedacted` / `AnyMapContainsRedacted` predicates and a frontend `hasRedactedConfig` predicate.
- **MSK Connect pagination.** `discoverMatchingConnectors` now pages `ListConnectors` via a `NextToken` loop (mirroring `ListClientVpcConnections`), so accounts with more than one page of connectors are no longer silently truncated. A mid-pagination error returns the connectors collected so far rather than discarding them.
- **No-wipe merge on re-run.** `AWSClientInformation.Connectors` was wholesale-replaced by the merge seam, so a denied or empty `ListConnectors` re-run overwrote prior connectors with nothing. A new `mergeConnectors` (dedup-by-name, new precedence, nil/empty short-circuit — mirroring `mergeSelfManagedConnectors`) preserves prior connectors on denial/empty and unions them on a successful overlapping re-run, through both `RefreshClusters` and `UpsertCluster`.

The redaction invariant is preserved throughout: the merge operates on already-redacted state, warnings are count-only, and a leak-free generated-Terraform test for each migrator asserts a redacted source config round-trips into the `.tf` as the placeholder (fail-closed), never as a working credential.

### Follow-up testing

- Full `go test ./...` — **50 packages ok, 0 failures**; `go vet` clean on all changed packages; frontend `tsc --noEmit` clean.
- Playwright e2e (real browser) asserts the UI banner is present with a redacted connector and absent without.
- New RED-first coverage for: pagination (multi-page / single-page / first-page-error / mid-pagination-error), merge (denial-preserves / overlap-dedup / edge cases / redaction-preserved), warning counts (positive / negative / partial "1 of 2"), and leak-free generated Terraform.

## Security considerations

Core guarantee verified: raw connector secrets never reach the state file or logs (redact-before-persist *and* redact-before-log; end-to-end tests assert no sentinel secret survives serialization). A dedicated security review of the follow-ups found **zero Critical/High issues** — the merge moves only already-redacted state, warnings are count-only, and the UI banner renders a constant with no value disclosure (no XSS).

Medium / needs-verification items, none blocking — all operator-controlled and matching pre-0.8.3 behavior:

1. **Cleartext basic-auth risk.** A scheme-less `--connect-rest-url` defaults to `http://`; with SASL/SCRAM the basic-auth credentials would then go over cleartext. Operators should pass `https://`.
2. **Credentials on argv.** The SASL password / TLS key flags are visible in process listings; mitigated by env-var binding (`utils.BindEnvToFlags`).
3. **Connector name in request path** is interpolated without `url.PathEscape`. Low risk given the same-host, operator-owned-cluster trust model; `url.PathEscape` would be a cheap future hardening.
4. **Translate-endpoint `baseURL` is test-only.** The follow-ups add an unexported `baseURL` field (defaulting to the hardcoded `https://api.confluent.cloud`) so tests can stub the translate API. It is intentionally **not** wired to any flag, opts field, or config. It should stay that way: the Cloud API key/secret are sent as a Basic-auth header to `baseURL`, so exposing it to user input would become an SSRF / credential-exfiltration vector.


adrian-januzi:
User-facing notice/output during scan indicating that some configs were redacted:
```shell
2026/06/23 11:30:18 INFO redacted sensitive connector config fields redacted_fields=2 connectors=2
2026/06/23 11:31:32 INFO Found existing state file, attempting to load it file=kcp-state.json
```

Example of the config in the state file:
```json
"connector_configuration": {
    "aws.access.key.id": "\u003ckcp-redacted\u003e",
    "aws.secret.access.key": "\u003ckcp-redacted\u003e",
    "connector.class": "io.confluent.connect.s3.S3SinkConnector",
    "flush.size": "75000",
    "format.class": "io.confluent.connect.s3.format.json.JsonFormat",
    "s3.bucket.name": "kcp-playground-s3-sink-connector",
    "s3.region": "us-east-1",
    "storage.class": "io.confluent.connect.s3.storage.S3Storage",
    "tasks.max": "1",
    "topics": "pageviews"
}
```

User notice that indicates configs contain redacted sensitive fields:

Screenshot 2026-06-23 at 11 41 43

`kcp create-asset migrate-connectors msk ...` command output w/ redacted fields warning:
```shell
Executing kcp with build version=0.0.0-localdev commit=58c85afc24c090a4c5006004bbd0b756c50e3458 date=2026-06-23T10:28:22Z
🔍 Found 2 connector(s) to migrate
⚠️  1 of 2 connector(s) contain redacted sensitive fields (<kcp-redacted>) — replace with real values in the generated Terraform before applying
✅ Successfully generated connector files for 2 connectors in kcp-playground-connectors
```

ifeeneyconfluent (approved):

ifeeneyconfluent (approved):

PR #356, opened 2026-06-23
feat(scan): restore self-managed-connector metrics collection

adrian-januzi:
## Summary

Restores self-managed **connector metrics** collection on `kcp scan self-managed-connectors`. The current branch restored connector *scanning* (with redaction) but not *metrics*, leaving `SelfManagedConnectors.Metrics` populated by nothing — PR #303 restores the UI that reads that field but had no producer feeding it. This PR re-wires the producer by porting the removed flags/credential-resolution/collection glue (recovered from `2aaddaaa`) onto the current restored scanner, mirroring the live `kcp scan clusters` metrics path. The metrics services, clients, and `ProcessedClusterMetrics` type were intact and are reused unchanged.

> Stacked on `feat/connector-scanning-redaction`. Base this PR on that branch; rebase if the base updates.

## What changed

- **Flags + validation (U1):** adds `--metrics <jolokia|prometheus>` plus `--metrics-duration`, `--metrics-interval` (default `10s`), `--metrics-range`, and `--credentials-file`. Validation mirrors `preRunScanClusters` — mutual exclusion via `Flags().Changed`, `duration > interval`, required-companion checks — and fails fast before any collection.
- **Collection + attachment (U2):** resolves the matching cluster's `jolokia`/`prometheus` credentials from the credentials file, collects Connect-worker metrics via the existing JMX/Prometheus services (`ConnectMetricDefinitions()` / `ConnectQueryDefinitions()`), and attaches them to `SelfManagedConnectors.Metrics`. Collection is **best-effort**: a failure warns and continues so already-scanned connectors always persist and the scan exits zero. MSK-only, matching the current restored scanner (OSK branch deferred — see plan Scope Boundaries).
- **Merge robustness (U3):** `mergeSelfManagedConnectors` now preserves a previously-collected metrics set across the zero-connector early-return branches, so re-running a scan without `--metrics` never wipes earlier metrics.

## Security posture (verified)

- **Credentials never logged or persisted (R11):** auth-section credentials flow only into request headers / TLS config, never into `kcp-state.json`; failure logs carry counts/error-class only. Covered by tests.
- **TLS honored, no silent downgrade (R12):** `WithJolokiaTLS`/`WithPrometheusTLS` applied whenever the creds carry a `tls` block. Covered by tests.
- **Untrusted endpoint responses non-fatal (R10):** malformed/unreachable/slow responses are caught and never crash or hang the scan (requests bounded by `ctx` + client timeouts). Covered by tests.

## Additional hardening: `connection.url` redaction

Closes a value-side secret leak in connector-config redaction (`internal/redact`). Redaction is key-based, so `connection.url` was preserved verbatim — but JDBC/Mongo connection strings routinely embed credentials in the value (e.g. `jdbc:mysql://user:pass@host/db`, `mongodb://user:pass@host/db`), which were then persisted to `kcp-state.json` and generated Terraform. `connection.url` is now in the static blacklist (with an explanatory comment + example), so its value is redacted wholesale. This also addresses the "endpoint URL with embedded credentials" reviewer note below for the connector-config path. Tests updated to assert redaction and lock the behavior in.

## Testing

Implemented test-first (RED→GREEN), including abuse cases (no-secret-in-logs, TLS-honored, ctx-no-hang, malformed/unreachable non-fatal) and an end-to-end re-run preservation test.

- `cmd/scan/self_managed_connectors`: 41 tests pass
- `internal/types` merge: 5 tests pass
- `internal/redact`: tests pass (`connection.url` redaction locked in)
- Full suite `go test ./...`: 50 packages ok, 0 failures · `go vet` clean · `go build ./...` clean

## Notes for reviewers (Medium, pre-existing — out of scope per plan)

These predate this change in the shared metrics clients/services and are flagged for awareness, not fixed here:

- **Unbounded response-body read** in the Jolokia/Prometheus clients (`io.ReadAll` with no `io.LimitReader` cap) — a hostile/buggy endpoint returning a very large body could exhaust memory. Pre-existing; freshly reachable via this command. The plan pre-acknowledges body-size limits as out of scope.
- **Endpoint URL persisted in `QueryInfo`** — leaks credentials only if an operator embeds them directly in the URL (`https://user:pass@host`) instead of the documented `auth:` section. Pre-existing shared-service transparency behavior. (The connector-config equivalent is now handled by the `connection.url` redaction above.)


adrian-januzi:
Retrieving metrics from our Connect cluster through JMX:
```shell
❯ jq '.msk_sources.regions[0].clusters[0].kafka_admin_client_information.self_managed_connectors.metrics.aggregates' kcp-state.json
{
  "connector-count": {
    "avg": 3,
    "max": 3,
    "min": 3,
    "p95": null,
    "p99": null
  },
  "source-record-poll-rate": {
    "avg": 10.033156573643272,
    "max": 10.848683467058425,
    "min": 9.30761152590163,
    "p95": null,
    "p99": null
  },
  "source-record-write-rate": {
    "avg": 10.039834668954526,
    "max": 10.776198656879588,
    "min": 9.349955476402492,
    "p95": null,
    "p99": null
  },
  "task-count": {
    "avg": 3,
    "max": 3,
    "min": 3,
    "p95": null,
    "p99": null
  }
}
```

ifeeneyconfluent (approved):

PR #358, opened 2026-06-23
feat(scan): Connect-specific metrics metadata

adrian-januzi:
## What

Self-managed Kafka Connect metrics (`SelfManagedConnectors.Metrics`) reused the broker-shaped `ProcessedClusterMetrics` / `MetricMetadata`. For a Connect cluster that meant the serialized `"metrics"` block carried ~9 empty broker fields plus an empty `region` / `cluster_arn` — pure noise, an artifact of type reuse rather than data we ever collect.

This replaces that with a purpose-built `ConnectClusterMetrics` / `ConnectMetricMetadata` carrying only the fields meaningful in the Connect scope: `start_date`, `end_date`, `period`, and `metrics_source`.

**Before**

```json
"metrics": {
  "region": "",
  "cluster_arn": "",
  "metadata": {
    "cluster_type": "", "number_of_broker_nodes": 0, "kafka_version": "",
    "broker_az_distribution": "", "enhanced_monitoring": "",
    "start_date": "...", "end_date": "...", "period": 300,
    "follower_fetching": false, "instance_type": "",
    "tiered_storage": false, "broker_type": ""
  }
}
```

**After**

```json
"metrics": {
  "metadata": {
    "start_date": "...",
    "end_date": "...",
    "period": 60,
    "metrics_source": "prometheus"
  },
  "results": [ ... ],
  "aggregates": { ... },
  "query_info": [ ... ]
}
```

## Testing

- Full Go suite green: **51 packages OK, 0 failed**. Frontend `tsc -b && vite build` clean.
- New tests (test-first): exact JSON-shape assertion (no broker fields), round-trip, per-backend `metrics_source` injection end-to-end, and metrics preservation across **both** the Set and merge paths.
- A type-level test confirms `ConnectClusterMetrics` *tolerates* legacy broker-shaped JSON (unknown broker fields ignored, `metrics_source` defaults empty). This is **not** a live state-file migration path, and shouldn't be read as one: the production loader (`NewStateFromBytes`) uses `DisallowUnknownFields` and would in fact reject a broker-shaped block. That's moot in practice — no released version (≤ v0.8.5) ever populated `SelfManagedConnectors.Metrics` (the `scan self-managed-connectors` command did not exist before this stack), so no state file in the wild carries broker-shaped Connect metrics and there is nothing real to migrate.

> Stacked on `feat/connector-metrics-restore` (PR base is that branch, not `main`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #360, opened 2026-06-24
feat(ui): show Connect metrics for MSK clusters, not just OSK

adrian-januzi:
## Summary

Kafka Connect is distribution-agnostic, but the Connect-metrics display added in #303 was wired OSK-only. MSK clusters already collect self-managed Connect metrics at scan time (onto `KafkaAdminClientInformation.SelfManagedConnectors.Metrics`) — they just weren't displayed. This generalizes the UI/API plumbing so the same charts/table/query view renders for MSK too, and hardens a few connector edges surfaced in review.

Stacked on top of #358 (`feat/connect-metrics-metadata`).

## Show Connect metrics for MSK clusters (`7556133d`)

- `FilterConnectMetrics` now takes a `sourceType` and searches only the matching source set (OSK by cluster ID, MSK by ARN) via per-source finder helpers, mirroring the `FilterClusterMetrics` dispatcher — a cluster id never resolves across source types.
- Replaced `GET /metrics/osk/:clusterId/connect` with a source-type route `GET /metrics/connect/:sourceType?clusterId=...`. The literal `connect` first segment avoids Echo static-node shadowing of `/metrics/:region/:cluster`, and `clusterId` moves to a query param so MSK ARNs (which contain `/`) aren't split across path segments.
- Frontend: `getOSKConnectMetrics` → `getConnectMetrics(sourceType, ...)`; `useConnectMetricsFetch` and `ConnectMetrics`/`ClusterConnectors` thread a required `sourceType`; `MSKClusterReport` passes the ARN + `"msk"`, `OSKClusterReport` passes `"osk"`.

## Hardening (`4e4380e1`)

- **migrate-connectors (self-managed):** guard the `connector.class` type assertion so a non-string value returns a per-connector error instead of panicking the whole run.
- **migrate-connectors (MSK):** added a path-traversal test asserting a hostile `../` connector name stays inside `OutputDir`.
- **ui/api + report:** distinguish "no Connect metrics ever collected" (404 + scan guidance via new `report.ErrNoConnectMetricsCollected` sentinel) from "collected, but none in the selected date range" (empty 200 → empty chart). Same disambiguation applied to the OSK broker-metrics handler.
- **discover:** match connectors to clusters by exact bootstrap `host:port` equality instead of substring containment.

## Testing

- `go test ./...` — all 51 packages pass.
- Tests cover OSK + MSK happy paths, date filtering, and abuse cases: cross-source bleed (both directions), unknown sourceType (4xx), and no-metrics (404).

### DIFF
--- a/cmd/create_asset/migrate_connectors/msk/msk_connectors_migrator_test.go
+++ b/cmd/create_asset/migrate_connectors/msk/msk_connectors_migrator_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/confluentinc/kcp/internal/redact"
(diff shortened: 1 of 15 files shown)


adrian-januzi:
https://github.com/user-attachments/assets/b47c3209-703a-4f00-8048-3f2a60b905ed

ifeeneyconfluent (approved):

PR #365, opened 2026-06-26
feat(state): backward-compatible kcp-state.json — migration engine, upgrade/version verbs, UI metadata + CI/archive guards

ifeeneyconfluent:
## Summary

Makes `kcp-state.json` **backward-compatible**: the current tool reads state files from **every release that produced one (v0.4.0 → v0.8.5)** — verified end-to-end against a real archive (32/32 load). Adds a versioned, migratable on-disk format with a migration engine, CI guards against accidental schema drift, graceful forward-failure, explicit `kcp state upgrade` / `kcp state version` verbs, and surfaces state-file metadata in the UI.

**Scope floor is `v0.4.0`** — the release that introduced `kcp-state.json` / `type State`. Pre-`v0.4.0` wrote a *different* file (`<region>-region-scan.json`); those and any unrelated JSON are treated as generic non-KCP files (no detection, no special message).

## What this delivers

**Reading old files (the core win)**
- Reads transparently **migrate in memory** before strict decode; originals are never mutated.
- Two real transformations recover the historical shapes: **B→C root reshape** (`regions` → `msk_sources`/`osk_sources`, v0.8.0) and **`schema_registries` array→object** (the v0.7.2 retype, recovering v0.4.2–v0.7.1). Everything else is additive/identity.
- Direction-accurate, dev-aware load errors (newer file → `kcp update`; unreadable → recreate — never circular `kcp state upgrade` advice). `kcp state upgrade` shares the loader and `SilenceUsage`s on runtime errors.

**Versioning & write path**
- `schema_version` is the single compatibility contract (build version demoted to provenance); `updated_at` + `migrated_from` breadcrumb stamped on write.
- A timestamped `.bak` is written before any migrating write.

**CLI**
- `kcp state upgrade --state-file <f>` — explicit, auditable migration to the current schema.
- `kcp state version --state-file <f>` — lenient raw-JSON metadata inspector; works on **any** file, even ones the strict loader rejects.

**UI** — a header info-popover showing the state-file metadata (schema version, KCP build, created/last-updated, migrated-from).

**Anti-regression / tests**
- Reflection **golden-schema snapshot guard** — any change to `State`'s shape fails CI until deliberately versioned (bump + upcaster + fixture + regenerate). *(CI-verified: a deliberate test PR with an improper schema change went red on this guard.)*
- **Golden-fixture load test** — Era C, latest Era B, and an array-form `schema_registries` fixture.
- **Opt-in S3-backed archive test** — loads all 32 real generated files v0.4.0–v0.8.5; local-only (not CI), triggered by `KCP_STATE_ARCHIVE`; the normal suite skips it.

## Notable decisions
- **v0.4.0 floor / Era A out of scope** (pre-v0.4.0 was a different file).
- **DF1 (per-section graceful degradation) descoped** — section-shape breaks are recovered by upcasters (full fidelity), not quarantine.
- The audit's predicted **v0.5.0 break was a false positive** (orphaned dead-code struct, never serialized); the real within-B break was the `schema_registries` retype, found by validating against the real archive.
- **Missing version stamp = unknown, not dev** so foreign/region-scan JSON get generic guidance. The superseded #308 "reject with running-version context" tests were reworked to the new contract.

## Testing
`go test ./...` green (53 pkgs, 0 failures; frontend built first per `CLAUDE.md`); golangci-lint clean; Playwright e2e for the UI popover; full v0.4.0–v0.8.5 archive loads 32/32. Design/spec lives outside the repo (per `CLAUDE.md`).

## Note for reviewers
The S3 archive (`s3://confluent-kcp-test-fixtures/state-archive/`, us-east-1) is local-dev-only and not reachable from CI by design — CI coverage comes from the committed golden-schema guard + golden-fixture load test (which includes the array-form case). ~16 commits; happy to walk through any area.


ifeeneyconfluent:
### Added: state-file metadata in the Web UI (commits `5c89afc`..`2a51d55`)

Building on the metadata fields this PR introduces (`schema_version`, `updated_at`, `migrated_from`), the UI now surfaces them so a user inspecting a loaded state file can see its provenance.

- **Backend:** widened the `ProcessedState` report DTO to carry `schema_version`/`updated_at`/`migrated_from` (alongside existing `kcp_build_info`/`timestamp`). `updated_at` is a `*time.Time` so a legacy file with no value is omitted from JSON (not serialized as `0001-01-01T00:00:00Z`).
- **Frontend:** a `StateMetadataPopover` (info icon in `AppHeader`) showing Schema version (`unversioned (legacy)` when 0), KCP build (version + commit/date), Created, Last updated, and Migrated from — empty/absent fields hidden.
- **Tests:** Go DTO unit tests (incl. the zero/legacy case) + a Playwright e2e (upload fixture → open popover → assert rows).

Display-only — no `kcp-state.json` schema change; the golden-schema guard stays green. Executed via subagent-driven development (per-task review + final whole-feature review, verdict: ready to merge).

ifeeneyconfluent:
## Ready for review — full feature summary

This PR makes `kcp-state.json` **backward-compatible**: the current tool reads state files from every release that produced one (**v0.4.0 → v0.8.5**), with a versioned, migratable format, CI guards against accidental schema drift, graceful forward-failure, and an explicit upgrade verb. It also surfaces state-file metadata in the UI and a new CLI command.

### Why
`kcp-state.json` is the on-disk projection of `types.State`; its shape changed repeatedly across releases (fields added/removed/retyped, the root reshaped). Since v0.8.1 the loader uses `DisallowUnknownFields`, so any older file failed to load with "regenerate/downgrade" advice that's often impossible or (for a newer file) plain wrong. This PR replaces that with a real schema-evolution story.

### What's included
**Migration engine & versioning** (`internal/state/migrate`, `internal/types/state.go`)
- `schema_version` as the single compatibility contract (build version demoted to provenance); `updated_at` + `migrated_from` breadcrumb.
- Migrate-on-read at the single read chokepoint (`NewStateFromBytes`): file-driven detection → ordered upcasters → strict decode. Reads never mutate disk.
- The two real transformations: **B→C root reshape** (`regions` → `msk_sources`/`osk_sources`, v0.8.0) and **`schema_registries` array→object** (v0.7.2 retype, recovering v0.4.2–v0.7.1).
- Write path stamps `schema_version`/`updated_at` and writes a timestamped `.bak` before any migrating write.
- Direction-accurate, dev-aware load errors (newer file → `kcp update`; unreadable → recreate — never circular `kcp state upgrade` advice).

**CLI**
- `kcp state upgrade --in <f> [--out <f>]` — explicit, auditable migration to the current schema.
- `kcp state version --in <f>` — lenient raw-JSON metadata inspector (schema version, build, timestamps, migrated-from); works on **any** file, even ones the strict loader rejects.

**UI** — a header info-popover showing the state-file metadata (schema version, KCP build, created/last-updated, migrated-from).

**Anti-regression / tests**
- Reflection golden-schema snapshot guard (schema drift fails CI until deliberately versioned).
- Golden-fixture load test (Era C, latest Era B, and an array-form `schema_registries` fixture).
- Opt-in S3-backed archive test that loads **all 32 real generated files v0.4.0–v0.8.5** — local-only (not CI), triggered by `KCP_STATE_ARCHIVE`; the normal suite skips it.

### Coverage achieved
Every real `kcp-state.json` from **v0.4.0** onward loads (verified 32/32 against a real archive). The audit's predicted v0.5.0 break turned out to be dead code (no-op); `schema_registries` array→object was the one genuine within-B retype.

### Scope decisions
- **Floor = v0.4.0** (the release that introduced `kcp-state.json`). Pre-v0.4.0 wrote a different file (`<region>-region-scan.json`) and is treated as unrelated JSON — out of scope by design.
- **Per-section graceful degradation (DF1) descoped** — section-shape breaks are recovered by upcasters (full fidelity), not quarantine.

### Testing
`go test ./...` green (frontend built first per `CLAUDE.md`); golangci-lint clean; Playwright e2e for the UI popover. The design/spec lives outside the repo (per `CLAUDE.md`).

### Note for reviewers
The S3 archive (`s3://confluent-kcp-test-fixtures/state-archive/`, us-east-1) is local-dev-only and not reachable from CI by design; CI coverage comes from the committed golden-schema guard + golden-fixture load test (which includes the array-form case).

internal/types/state.go
```diff
@@ -156,6 +193,37 @@ func (s *State) WriteToFile(filePath string) error {
...
+		return nil // same version → not a migrating write
+	}
+	ts := time.Now().UTC().Format("20060102T150405Z")
+	bak := fmt.Sprintf("%s.%s.bak", filePath, ts)
+	for i := 1; ; i++ {
+		if _, err := os.Stat(bak); os.IsNotExist(err) {
+			break
+		}
+		bak = fmt.Sprintf("%s.%s-%d.bak", filePath, ts, i)
+	}
```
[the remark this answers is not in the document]
ifeeneyconfluent: Good catch — fixed in 8dc5c56. \`backupIfMigrating\` now treats only \`os.IsNotExist\` as "nothing to back up"; any other read error aborts the migrating write rather than silently overwriting an unreadable file with no backup (preserving the D7 guarantee). The \`.bak\` collision loop now breaks on any non-nil \`Stat\` error, removing the possible infinite loop on a non-NotExist stat error. Added a (root-guarded) test for the unreadable-file case.

internal/types/state.go
```diff
@@ -81,39 +86,71 @@ func NewStateFromFile(stateFile string) (*State, error) {
...
-				if strings.HasSuffix(raw.KcpBuildInfo.Version, "-localdev") {
-					return nil, fmt.Errorf("%v (file was created with a local dev build %q, you are running %q). Please recreate the state file with kcp discover (MSK) or kcp scan clusters (Apache Kafka)", err, raw.KcpBuildInfo.Version, build_info.Version)
-				}
-				return nil, fmt.Errorf("%v (file was created with KCP version %q, you are running %q). Please recreate the state file with kcp discover (MSK) or kcp scan clusters (Apache Kafka) using the latest KCP release, or use KCP version %s to load this file", err, raw.KcpBuildInfo.Version, build_info.Version, raw.KcpBuildInfo.Version)
-			}
-			return nil, fmt.Errorf("%v. Please recreate the state file with kcp discover (MSK) or kcp scan clusters (Apache Kafka) using the latest KCP release", err)
-		}
-		return nil, fmt.Errorf("%v. Please recreate the state file with kcp discover (MSK) or kcp scan clusters (Apache Kafka) using the latest KCP release", err)
+		return nil, fmt.Errorf("state file did not match the current schema after migration (migrated from %s): %w. Run `kcp state upgrade --in <file>`, or recreate it with `kcp discover` / `kcp scan clusters`", fromLabel, err)
```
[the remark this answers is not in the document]
ifeeneyconfluent: Agreed, and already addressed in 565d32c (before this comment). The decode-failure message no longer suggests \`kcp state upgrade\` (which shares this exact loader and would fail identically); it now gives generic recreate guidance: "…recreate it with \`kcp discover\` / \`kcp scan clusters\`, or use a kcp release that supports it." Nice independent corroboration of that fix.

internal/types/state.go
```diff
@@ -81,39 +86,71 @@ func NewStateFromFile(stateFile string) (*State, error) {
...
 	}
 
+	// A dev-stamped file that decodes cleanly is a clean success — no warning
```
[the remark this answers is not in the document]
ifeeneyconfluent: Verified this is already enforced: \`migrate.Upgrade\` runs \`detectVersion\` first, which uses \`json.Unmarshal\` (strict on trailing data) before the decoder is reached — so \`{…}{…}\` is rejected at inspection, not silently accepted. Added a regression test (\`TestNewStateFromBytes_TrailingDataRejected\`) in 8dc5c56 to lock the contract rather than add a redundant decode-site check.

internal/types/state_test.go
```diff
@@ -1152,28 +1163,30 @@ func TestNewStateFromBytes_EmptyVersion(t *testing.T) {
 	}
 }
 
-func TestNewStateFromBytes_UnknownFields_WithVersion_RejectsWithVersionContext(t *testing.T) {
+// Legacy Era B files (top-level `regions`) are recognised by the migration engine but
+// have no upcaster in Plan 1 (the B→C step lands in Task 7). The loader must surface the
+// actionable `kcp state upgrade` guidance rather than the superseded #308 "recreate or
+// downgrade with version context" message. The "kcp state upgrade" guidance is stable
+// across Task 3 (ErrUnsupportedLegacy) and Task 7 (post-B→C decode guidance).
```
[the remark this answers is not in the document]
ifeeneyconfluent: Already fixed in 565d32c. That comment was reworked; the current one attributes the failure to a field-level mismatch (e.g. \`region_name\` vs \`name\`) after the B→C reshape — exactly as you note — not the absence of upcasting.

cmd/state/upgrade/cmd_state_upgrade.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/confluentinc/kcp/internal/types"
+	"github.com/spf13/cobra"
+)
+
+func NewStateUpgradeCmd() *cobra.Command {
+	var in, out string
+	cmd := &cobra.Command{
+		Use:           "upgrade",
+		Short:         "Migrate a kcp-state.json file to the current schema",
+		Long:          "Reads a state file produced by any prior KCP version, migrates it to the current schema, and writes the result. Writes in place if --out is omitted.",
```
adrian-januzi: Thoughts on having an `Example:` section here for the docs? I'm on the fence about it personally because on one hand it is consistent but on the other hand its a super straight forward command 🤷

internal/state/migrate/migrate.go
```diff
@@ -0,0 +1,91 @@
...
+			label = "kcp_build_info.version=" + buildVersion
+		}
+		return data, label, nil
+	}
+
+	// Legacy file: run the ordered upcaster chain.
+	var doc map[string]any
+	if err := json.Unmarshal(data, &doc); err != nil {
+		return nil, "", fmt.Errorf("failed to parse legacy state file: %w", err)
+	}
```
adrian-januzi, line 77: From what I understand there is a chance here that during the unmarshal, we could in-theory lose some precision as `json.Unmarshal()` converts any number to a Float.

> To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:
> ...
> float64, for JSON numbers
source: https://pkg.go.dev/encoding/json#Unmarshal

I checked our state files and for our testing ones, we don't actually hit this scenario and Claude agrees, so I am in favour of ignore it but I thought I would raise it anyway.

.gitignore
adrian-januzi, line 1: Can we add something like `*.bak` or `kcp-state.json*.bak` (I think that might work) so that we (me) don't accidentally commit backups to the repo :)

Makefile
```diff
@@ -126,6 +126,37 @@ test-schema-registry: build ## Run Schema Registry scan tests (unauthenticated,
...
+		if command -v shasum >/dev/null 2>&1; then echo "$$CHECK" | shasum -a 256 -c -; \
+		else echo "$$CHECK" | sha256sum -c -; fi
+	@tar -xzf $(STATE_ARCHIVE_CACHE)/state-archive.tar.gz -C $(STATE_ARCHIVE_CACHE)
+	@echo "state archive ready at $(STATE_ARCHIVE_CACHE) ($$(ls -d $(STATE_ARCHIVE_CACHE)/v* | wc -l | tr -d ' ') versions)"
+
+# No build-frontend dep: internal/types does not embed the frontend dist, so the load test
+# runs without it (keeps this S3-backed target fast).
+test-state-archive: fetch-state-archive ## Load every archived real state file through the loader (opt-in regression test)
+	KCP_STATE_ARCHIVE=$(abspath $(STATE_ARCHIVE_CACHE)) go test ./internal/types/ -run TestStateArchiveLoads -v
+
```
adrian-januzi, line 159: Lovely!

adrian-januzi (commented):
few non-blocking nits. Tested some happy paths but will test a bit more chaotic stuff after lunch 🫡

internal/state/migrate/migrate.go
ifeeneyconfluent: Good catch — you're right that `json.Unmarshal` into `map[string]any` widens every JSON number to `float64`, which would lose integer precision above 2^53 and re-serialize values ≥ 1e21 in scientific notation (e.g. `1e+21`), which the strict `types.State` decode then rejects.

As you found, no real archived file hits this (the v0.4.0–v0.7.1 corpus tops out at `86400`), and the upcasters never read numeric values — they only reshuffle top-level keys. But since the round-trip should be faithful regardless of magnitude, I've hardened the decode to use `json.Decoder` + `UseNumber()`, so every number survives as its exact `json.Number` literal. Verified all 32 archived files (v0.4.0–v0.8.5) still load via the state-archive test.

Done in 5a049ab.

.gitignore
ifeeneyconfluent: Good idea. Added `*.bak` to `.gitignore`.

Worth noting the existing `*.json` rule didn't already cover these: `backupIfMigrating` names backups `kcp-state.json.<ts>.bak` (and a `-<n>` collision variant), so they end in `.bak`, not `.json`. `*.bak` catches both. Done in e3366ea.

adrian-januzi (approved):
Feel like I've tested it robustly and couldn't get it to not upgrade/fail mid-upgrade unless I made manual changes to the state file.

The only nitpick I have left is around the `--in` and `--out` flags. Not sure how I feel about them as we typically have `--state-file` throughout the codebase when asking for the state file and something like `--output-X` for user-named [output files](https://github.com/confluentinc/kcp/blob/main/cmd/create_asset/migrate_connectors/msk/cmd_migrate_msk_connectors.go#L65).

Not a blocker and I don't mind either way.

cmd/state/upgrade/cmd_state_upgrade.go
ifeeneyconfluent: Added an `Example:` here — and, for symmetry, to `kcp state version` too (so the `state` subtree stays consistent; adding it to only one would have been the inconsistency you were weighing).

While in here I also reworked this command's flags (per a separate request): `--in` → `--state-file` to match the rest of the CLI (`ui`, `scan`, `report`, `healthcheck`), and dropped `--out` so `upgrade` always writes in place — `backupIfMigrating` already preserves the original as `<state-file>.<UTC-timestamp>.bak`, which the help text and example now call out. Done in 7ccbd14.

ifeeneyconfluent:
@adrian-januzi re: your approval nitpick about the `--in`/`--out` flags on `kcp state` — good call, addressed in 7ccbd14:

- **`--in` → `--state-file`** on both `kcp state version` and `kcp state upgrade`, matching the convention used everywhere else (`ui`, `scan`, `report`, `healthcheck`).
- **Dropped `--out`** entirely. `upgrade` now always writes in place; `backupIfMigrating` already preserves the original as `<state-file>.<UTC-timestamp>.bak`, so there was no real need for a user-named output path. The command's `Long`/`Example` and CLAUDE.md now call out the `.bak` behaviour.

Thanks for the thorough testing 🙏

adrian-januzi (approved):
lgtm

adrian-januzi (approved):
lgtm take two

PR #366, opened 2026-06-26
fix(scan): clearer schema-registry URL errors + Apicurio guidance

adrian-januzi:
## What & why

Improves the experience of running `kcp scan schema-registry --sr-type confluent` against Confluent-compatible registries (notably **Apicurio**).

Previously, pointing `--url` at the wrong place — an Apicurio host root, an ingress/route, a UI, or a login proxy — failed with an opaque JSON-decode error:

```
ERROR failed to scan schema registry: failed to get default compatibility:
      invalid character '<' looking for beginning of value
```

The `<` is the first byte of an HTML page: the confluent-kafka-go client decodes any 2xx body as JSON without checking `Content-Type`, so an HTML response surfaces as a cryptic parse failure rather than "you pointed at the wrong endpoint."

## Changes

1. **Pre-flight URL validation** (`internal/client/schema_registry_preflight.go`)
   Before scanning, `GET <url>/config` (the same endpoint the scan's first call hits) and return a clear, class-distinct error when the endpoint is not a usable Schema Registry:
   - non-JSON / HTML body → "does not look like a Schema Registry REST API" (status + content-type, **no** body dump)
   - `3xx` redirect → refused (a real SR responds directly), reports the `Location`
   - `401/403` → authentication error, points at `--use-basic-auth` / `--username` / `--password`
   - non-2xx with a JSON error body → surfaces the registry's own `{"error_code","message"}`
   - Redirects are refused (SSRF guard); response snippets are sanitized and length-bounded; credentials are never echoed.

2. **`--sr-type` help text** (`cmd/scan/schema_registry/cmd_scan_schema_registry.go`)
   Notes that `confluent` works against any registry exposing a Confluent-compatible REST API — e.g. Apicurio via its `/apis/ccompat/v7` endpoint — the most common gotcha when the target isn't Confluent Schema Registry.

## Testing

- Unit tests (`schema_registry_preflight_test.go`, `cmd_scan_schema_registry_test.go`) cover: HTML on 2xx, a lying `Content-Type: application/json` with an HTML body, `401`/`403`, a JSON error body on non-2xx, HTML on non-2xx, redirects, and an unreachable host. All assert the raw `invalid character '<'` decode error never leaks and no response body / credentials are dumped.
- `go test ./cmd/scan/schema_registry/ ./internal/client/` passes.
- Validated end-to-end against a live `apicurio-registry-mem:2.6.5.Final`: scanning the bare host now returns a clear error, and scanning the `/apis/ccompat/v7` endpoint discovers all subjects/versions cleanly.

### DIFF
--- a/cmd/scan/schema_registry/cmd_scan_schema_registry.go
+++ b/cmd/scan/schema_registry/cmd_scan_schema_registry.go
@@ -72,7 +72,7 @@ func NewScanSchemaRegistryCmd() *cobra.Command {
 	requiredFlags := pflag.NewFlagSet("required", pflag.ExitOnError)
 	requiredFlags.SortFlags = false
 	requiredFlags.StringVar(&stateFile, "state-file", "", "The path to the kcp state file.")
-	requiredFlags.StringVar(&srType, "sr-type", "", "Schema registry type: 'confluent' or 'glue'")
+	requiredFlags.StringVar(&srType, "sr-type", "", "Schema registry type: 'confluent' or 'glue'. Use 'confluent' for any registry exposing a Confluent-compatible REST API, e.g. Apicurio via its /apis/ccompat/v7 endpoint.")
 	schemaRegistryCmd.Flags().AddFlagSet(requiredFlags)
 
 	confluentFlags := pflag.NewFlagSet("confluent", pflag.ExitOnError)
--- a/cmd/scan/schema_registry/cmd_scan_schema_registry.go
+++ b/cmd/scan/schema_registry/cmd_scan_schema_registry.go
@@ -179,6 +179,14 @@ func runScanConfluentSchemaRegistry() error {
 		return fmt.Errorf("failed to get auth option: %v", err)
 	}
 
+	// Pre-flight the endpoint before handing off to the confluent-kafka-go client,
+	// which would otherwise surface a non-JSON (HTML) response as the opaque
+	// "invalid character '<'". The validator's error is already actionable, so it
+	// is returned as-is rather than re-wrapped.
+	if err := client.ValidateConfluentSchemaRegistryURL(opts.Url, authOption); err != nil {
+		return err
+	}
+
 	schemaRegistryClient, err := client.NewSchemaRegistryClient(opts.Url, authOption)
 	if err != nil {
 		return fmt.Errorf("failed to create schema registry client: %v", err)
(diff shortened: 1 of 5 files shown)


adrian-januzi:
## ✅ Verified against a live Apicurio registry

Built this branch and ran it against a real `apicurio-registry-mem:2.6.5.Final` (the in-memory Apicurio image exposing the Confluent-compatible `/apis/ccompat/v7` API). This reproduces — and resolves — the original report, where `kcp scan schema-registry --sr-type confluent --url https://apicurio.eventstreams.io ...` failed on `0.8.5` with:

```
ERROR failed to scan schema registry: failed to get default compatibility:
      invalid character '<' looking for beginning of value
```

### 1. Correct ccompat URL now scans cleanly

```
kcp scan schema-registry --sr-type confluent \
  --url http://127.0.0.1:8080/apis/ccompat/v7 --use-unauthenticated --state-file ...
→ ✅ Successfully scanned schema registry
```

The discovered registry was written to state in full — confirming kcp **is** compatible with Apicurio via the ccompat endpoint:

| | |
|---|---|
| Subjects | **5** |
| Versions | **8** (3 subjects have v1 + v2 — versioned schemas round-trip correctly) |
| Schema types | **AVRO ×5, JSON ×1, PROTOBUF ×2** |
| Compatibility | per-subject BACKWARD / FULL / FORWARD / NONE captured; default BACKWARD |

### 2. The opaque decode error is gone — replaced by an actionable one

The original failure came from the host returning an **HTML** page (from an ingress/proxy) where kcp expected JSON. Reproduced against an HTML-returning host:

> **Before (0.8.5):** `failed to get default compatibility: invalid character '<' looking for beginning of value`
>
> **After (this PR):** `<host> does not look like a Schema Registry REST API: it returned HTTP 404 with a non-JSON (text/html) body; verify --url points to the Schema Registry endpoint`

Pointing at a host that returns a JSON error (e.g. the bare Apicurio host, no ccompat path) now surfaces the registry's own message instead of choking:

> `schema registry at <host> returned HTTP 404: {"...RESTEASY003210: Could not find resource for full path: .../config"...}`

### Takeaway for users

The fix for the original report is to point `--url` at the Confluent-compat base — `https://<host>/apis/ccompat/v7` instead of the bare host. With this PR, a wrong URL produces a message that says exactly that, instead of a cryptic JSON parse error. (If the endpoint is fronted by an OAuth/SSO proxy rather than HTTP Basic, the pre-flight now reports a clean `401` — kcp can only authenticate via unauthenticated or HTTP Basic.)

ifeeneyconfluent (approved):

PR #371, opened 2026-07-02
fix(migration): confirm mirror topics reach STOPPED before switchover; add optional promote batching

tjunderhill:
## Problem

`kcp migration execute` could report success and switch the gateway while some mirror topics were still in `PENDING_STOPPED` on the destination.

Root cause: `PromoteTopics` treated a `mirrors:promote` response of `error_code: 0` as "promotion complete" and immediately moved on. But promote is asynchronous — `error_code: 0` only means the request was *accepted/enqueued*, not that mirroring has actually stopped. KCP never confirmed the terminal `STOPPED` state, so it advanced to the gateway switchover while the backend was still draining promotions.

This is most visible on migrations with large topic counts, where the whole set is submitted in a single promote call and the backend's asynchronous promotion has not completed by the time KCP switches over.

## Changes

**Correctness fix** (`internal/services/migration/workflow.go`)
- After a promote is accepted, the topic is held in an `awaitingStop` set rather than treated as done.
- Each poll re-reads `ListMirrorTopics` and only removes a topic from the work set once its `mirror_status == STOPPED`.
- `PromoteTopics` does not return — and the workflow does not proceed to `SwitchGateway` — until **every** topic is confirmed `STOPPED`. Uses a positive check for the terminal state, so any transient/unknown status is safely treated as "not done yet." Per-topic promote-error retry logic is unchanged.

**New optional flag** `--promote-batch-size N` (`cmd/migration/execute`)
- Default `0` = promote all caught-up topics at once (existing behaviour).
- When `>0`, promotes in synchronous batches: promote `N`, wait for all `N` to reach `STOPPED`, then the next batch. Wired through `MigrationExecutorOpts` → `workflow.SetPromoteBatchSize()`, mirroring `--rollout-timeout`.

**Tests**
- Unit: stuck-`PENDING_STOPPED`-does-not-succeed, waits-for-`STOPPED`, and synchronous-batch behaviour (`[10,10,5]`, no cross-batch overlap); plus `--promote-batch-size` flag parsing (default/explicit/invalid/env). Existing workflow + orchestrator tests updated to model the `ACTIVE → STOPPED` transition.
- E2E (Minikube + CFK): baseline now asserts the mirror reaches `STOPPED` after execute; new `TestMigrationE2E_PromoteBatchSize` provisions a 10-topic cluster link, runs `execute --promote-batch-size 1`, and asserts `switched` + all 10 mirrors `STOPPED` + (from `kcp.log`) 10 single-topic promotes with zero bulk calls. `setup.sh` gained a `batch` scenario and multi-topic provisioning; the single-topic `cluster-link.yaml` template was replaced by an in-shell generator that produces identical output for existing scenarios.

**Docs**
- `getting-started-with-zero-cut-migrations.md`: documented `--promote-batch-size` (optional, default `0`) and clarified that promote confirms each topic reaches `STOPPED` before switching.

## Testing

- `go test ./internal/services/migration/... ./internal/services/clusterlink/... ./cmd/migration/...` — pass
- `make test-migration` (full Minikube + CFK lifecycle) — pass, all 5 E2E tests green, clean teardown
- `go vet`, `gofmt`, `go build ./...` — clean

## Notes

- No promote-timeout flag: the `STOPPED` wait has no deadline (consistent with the existing `--rollout-timeout`/lag-wait no-deadline default). Aborting is Ctrl-C; the migration is resumable from persisted FSM state.


cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -90,6 +91,7 @@ the migration state file and must be provided each time.`,
 	optionalFlags.SortFlags = false
 	optionalFlags.BoolVar(&insecureSkipTLSVerify, "insecure-skip-tls-verify", false, "Skip TLS certificate verification for REST endpoint and Kafka connections.")
 	optionalFlags.DurationVar(&rolloutTimeout, "rollout-timeout", 0, "Maximum time to wait for the Confluent operator to report the gateway as Ready during fence and switchover. 0 (the default) means no deadline — the wait runs until the operator converges or the user cancels.")
+	optionalFlags.IntVar(&promoteBatchSize, "promote-batch-size", 0, "Maximum number of mirror topics to promote per batch. 0 (the default) promotes all topics at once. When set (>0), each batch is promoted and confirmed STOPPED before the next batch is submitted.")
```
ifeeneyconfluent, line 94: Is 1 batch with all topics safe for large topic counts?  Are there any upper limits to the batch size of topics we can request be promoted?

ifeeneyconfluent: In general, why do we offer this?  If there is no upper limit, why would Omar want to batch? How would he know what batch size to use?

internal/services/migration/workflow.go
```diff
@@ -327,14 +345,60 @@ func (s *MigrationWorkflow) PromoteTopics(ctx context.Context, config *Migration
...
+				status := statusByTopic[topic]
+				if status == clusterlink.MirrorStatusStopped {
+					fmt.Printf("   %s %s stopped\n", color.GreenString("✔"), topic)
+					slog.Debug("mirror topic promotion confirmed stopped", "topic", topic)
+					delete(awaitingStop, topic)
+					delete(remaining, topic)
+				} else {
+					slog.Debug("mirror topic promotion still pending",
+						"topic", topic, "status", status)
+				}
```
adrian-januzi, line 370: Is it possible that a network or authentication issue (like shown in the docs here: https://docs.confluent.io/platform/current/multi-dc-deployments/cluster-linking/mirror-topics-cp.html#view-mirror-topic-state-transition-errors) might cause kcp to poll indefinitely?

internal/services/migration/workflow.go
```diff
@@ -327,14 +345,60 @@ func (s *MigrationWorkflow) PromoteTopics(ctx context.Context, config *Migration
 		default:
 		}
 
+		// Confirm accepted promotions have actually reached STOPPED. Until a
+		// topic is verified STOPPED it stays in `remaining`, which keeps the
+		// workflow in the promote phase and blocks the gateway switchover.
+		if len(awaitingStop) > 0 {
```
adrian-januzi, line 351: If there is a failure or the switchover is stopped mid-polling for topics, we would lose the `awaitingStop` map of topics. On a restart, the `awaitingStop` will be empty so we would move immediately through `len(awaitingStop) > 0`.

adrian-januzi: Claude:
> Resume after mid-promote interruption re-promotes already-STOPPED topics and can wedge execute with the gateway fenced

```
The bug: resume starts blind

On a re-run, PromoteTopics seeds its worklist purely from config (workflow.go:334-339):

remaining := make(map[string]bool)
awaitingStop := make(map[string]bool)
for _, topic := range config.Topics {
    remaining[topic] = true
}
```

cmd/migration/execute/cmd_migration_execute.go
adrian-januzi: On the other hand, what happens if we enter `-1`, `-100`?

adrian-januzi (approved):

PR #367, opened 2026-06-26
feat(migration): add --detect-unrouted-producers flag to migration execute

bmartin494:
## Summary

- Adds required `--detect-unrouted-producers-duration` flag to `kcp migration execute` that detects producers bypassing the gateway and writing directly to the source cluster after fencing
- Detection is modelled as its own FSM step: `verify_fence` (`fenced → fence_verified`) takes two source-offset snapshots separated by the user-specified duration. Any partition with an increasing offset — including partitions that appear between snapshots, checked against a zero baseline — means a producer is bypassing the fence
- On detection, the FSM rolls back `fenced → initialized` via the `abort_fence` transition, whose action unfences the gateway (initial CR reapplied with server metadata stripped) and **waits for gateway readiness** before reporting traffic restored
- If the unfence apply or its rollout fails, the `abort_fence` transition is cancelled — persisted state stays `fenced`, matching reality, and re-running execute retries
- `fence_verified` is a point-in-time attestation and never survives a restart: FSM bootstrap fires an explicit `expire_verification` transition (`fence_verified → fenced`) so a resume re-runs the detection window — a rogue producer appearing between a failed promote and a re-run hours later is still caught
- Network/fetch errors during the monitoring window are distinguished from detections: they propagate without triggering unfence or rollback (state stays `fenced`; re-run retries verification)
- Value of `0` skips the check with an explicit "Detection disabled — skipping check" line. Non-zero values must be at least `10s`. The flag is required to force operators to explicitly consider the monitoring window
- Also: user-facing migration output routed through a single reporter seam; action callbacks registered per-event; `SilenceUsage` suppresses help text on execute errors

## FSM

```mermaid
stateDiagram-v2
    [*] --> uninitialized
    uninitialized --> initialized: initialize
    initialized --> lags_ok: wait_for_lags
    lags_ok --> fenced: fence
    fenced --> fence_verified: verify_fence
    fenced --> initialized: abort_fence
    fence_verified --> fenced: expire_verification
    fence_verified --> promoted: promote
    promoted --> switched: switch
```

`abort_fence` fires only when `verify_fence` detects unrouted producers; `expire_verification` fires only at FSM bootstrap. A promote failure rests at `fence_verified` (the true mid-run position); the next run demotes it and re-verifies.

## Background

[...]

PR #380, opened 2026-07-09
fix(migration): make --detect-unrouted-producers-duration optional and hide it (WIP)

bmartin494:
## Summary

The unrouted (rogue) producer detection added in #367 required the `--detect-unrouted-producers-duration` flag on `kcp migration execute`. In practice, traffic that is slow but legitimate can keep arriving despite the fence and trip a false positive, blocking the user from completing the migration.

This PR steps the feature back to experimental/WIP without ripping it out:

1. **Makes the flag optional**, defaulting to `0` (skip). Users are no longer forced to use it.
2. **Hides the flag from `--help`** (`MarkHidden`). It is no longer advertised as a supported option — but it **still parses and the detection still runs when explicitly set**, so the capability remains available for internal testing.
3. **Stops referencing the flag in the getting-started guide** so public docs match the "not advertised" state.

The detection code path itself is unchanged and stays in place while the false-positive edge case is worked out.

> Note: "hidden" means removed from `--help` only — the flag still exists in source and remains functional if typed. It is not gated or removed.

## Changes

- Move `--detect-unrouted-producers-duration` from the required flag set to the optional flag set and drop the `MarkFlagRequired` call.
- `MarkHidden` the flag so it does not appear in `kcp migration execute --help`.
- Label the flag `[Experimental]` in its (now hidden) help text and note the false-positive caveat.
- Rework the getting-started guide's "Verify fence" phase to describe a default pass-through and drop the flag name.
- Replace the `..._Required` test with `..._OptionalDefaultsToSkip`, pinning that omitting the flag proceeds (defaults to `0`/skip) rather than failing validation. The `>= 10s` minimum still applies when a non-zero value is passed.

## Testing

- `go test ./cmd/migration/execute/...` passes.
- Verified `kcp migration execute --help` no longer lists the flag.
- Verified passing `--detect-unrouted-producers-duration 10s` still parses and runs (reaches the state-file error, not an "unknown flag" error).


docs/assets/getting-started-with-zero-cut-migrations.md
```diff
@@ -250,7 +250,7 @@ Performs the cutover in six automatic phases. The operation is resumable: if int
...
 | **Block**             | Applies the fenced CR to the gateway; the route stops accepting produce/consume requests                                                         | `BROKER_NOT_AVAILABLE`; standard clients buffer and retry automatically |
 | **Pause offset sync** | With `--pause-consumer-offset-sync`, pauses cluster-link consumer offset sync so destination consumer offsets freeze at their freshest values; skipped otherwise | Still retrying                                                          |
-| **Verify fence**      | Monitors source offsets for `--detect-unrouted-producers-duration` to catch producers bypassing the gateway; on detection, unblocks and restores offset sync automatically | Still retrying                                                          |
+| **Verify fence**      | Currently a pass-through — KCP proceeds straight to promotion. (An internal, experimental check that monitors source offsets for producers bypassing the gateway exists but is disabled by default while a false-positive edge case is worked out.) | Still retrying                                                          |
```
tjunderhill: Can we drop this

bmartin494: Removed in latest commit

tjunderhill (approved):
LGTM!

adrian-januzi (approved):

PR #381, opened 2026-07-09
feat(migration): add --consumer-offset-sync-drain-duration to reduce duplicate processing

tjunderhill:
## Summary

Adds an optional `--consumer-offset-sync-drain-duration` flag to `kcp migration execute` that reduces duplicate message processing across a cutover when `--pause-consumer-offset-sync` is in use.

## Background

When a migration is initialised with `--pause-consumer-offset-sync`, the `pause_offset_sync` FSM stage disables the cluster link's `consumer.offset.sync.enable` immediately after the gateway is fenced. Consumer offset sync runs on an interval (`consumer.offset.sync.ms`), so at fence time the destination's copy of each consumer group's committed offset can trail the source by up to one interval. If sync is disabled the instant the fence lands, those final commits never propagate to the destination, and consumers reprocess them after switchover.

The fence freezes source consumer offsets — clients can no longer commit — so there is a safe window after fencing in which the link can still propagate the final committed offsets before sync is turned off.

## What this adds

`--consumer-offset-sync-drain-duration` (optional, default `0`): when `> 0`, the `pause_offset_sync` stage holds after fencing — with sync still enabled — for the given duration before disabling `consumer.offset.sync.enable`. This lets the cluster link run one or more further sync cycles and land the final committed offsets on the destination, shrinking the set of messages reprocessed after switchover.

- Default `0` = no wait = the prior immediate-disable behaviour (fully backward compatible).
- Has no effect unless the migration was initialised with `--pause-consumer-offset-sync` (the pause stage is otherwise a no-op).
- **Best-effort:** offset sync is asynchronous, so the drain reduces but does not guarantee zero duplicates, and it does not cover messages a consumer processed but had not yet committed on the source before the fence. The flag help text and docs state this explicitly.
- The only cost is that the gateway stays fenced (clients buffering/retrying) for the drain duration.

## Changes

**Behaviour** (`internal/services/migration/workflow.go`)
- `PauseOffsetSync` gains a drain window: after the drift check confirms `consumer.offset.sync.enable=true` and before the disable, it waits `ConsumerOffsetSyncDrainDuration` (ctx-cancellable) when `> 0`.
- The drift check still runs first, so a misconfigured link fails fast without waiting.
- A ctx cancellation during the drain leaves sync still enabled (nothing flipped), consistent with the existing drift-refusal path.

**Config** (`internal/services/migration/state.go`)
- New `ConsumerOffsetSyncDrainDuration` field on `MigrationConfig`, applied per-run from the flag, mirroring `DetectUnroutedProducersDuration`.

**Flag** (`cmd/migration/execute`)
- `--consumer-offset-sync-drain-duration` (optional, default `0`); negative values rejected at validation.

**Docs** (`getting-started-with-zero-cut-migrations.md`)
- Documented the flag and the phase-table behaviour, framed as best-effort.

**Tests**
- Unit: drain holds before disabling; ctx cancellation during the drain leaves sync enabled and flips nothing; `0` disables immediately; flag parsing (default / explicit / negative rejected / env var).
- E2E (`TestMigrationE2E_PauseOffsetSync_Drain`, dedicated `pause-sync-drain` scenario): with `--pause-consumer-offset-sync` and `--consumer-offset-sync-drain-duration 15s`, asserts the migration completes to `switched`, stdout is ordered fence → pause → drain → disable, execute wall-clock is at least the drain duration, and the cluster-link config is restored to `true`. This proves the drain path runs end-to-end against a real cluster link and does not regress the offset-sync round-trip; it does not attempt to prove a statistical reduction in duplicates, which is timing-dependent and not deterministically observable in the harness.

## Testing

- `go test ./internal/services/migration/... ./cmd/migration/...` — pass
- `go build ./...`, `go vet`, `gofmt` — clean
- Migration E2E against Minikube + CFK: `TestMigrationE2E_PauseOffsetSync_Drain` — PASS (~141s, all subtests)


integration-tests/migration/migration_e2e_test.go
```diff
@@ -1055,6 +1056,140 @@ func TestMigrationE2E_PauseOffsetSync_HappyPath(t *testing.T) {
...
+// TestMigrationE2E_PauseOffsetSync_HappyPath). The deterministic evidence here
+// is the ordered stdout (fence → pause → drain → disable) plus a wall-clock
+// lower bound.
+//
+// Runs against the dedicated "pause-sync-drain" scenario provisioned by
+// setup.sh, so it is independent of the other pause-sync tests.
+func TestMigrationE2E_PauseOffsetSync_Drain(t *testing.T) {
+	cfg := loadEnvConfig(t, scenarioPauseSyncDrain)
+
+	stateFile := "/workspace/migration-state-pause-sync-drain.json"
```
adrian-januzi, line 1079: What's this `workspace/` dir? Is this a temporary/on-the-fly generated directory and file?

tjunderhill: What it is

kcp in these tests runs inside the kcp-runner pod — runKCP does kubectl exec … /workspace/kcp <args> (line 153-154). So /workspace/… is a path inside that pod, not on your Mac. /workspace is where setup.sh copies the kcp binary and fixtures.

migration-state-pause-sync-drain.json is the migration's state file — the persisted FSM state and config for one migration (the MigrationState/MigrationConfig we touched in state.go). It's the migration-command equivalent of kcp-state.json.

How the test uses it

It's threaded through every kcp invocation via --migration-state-file, and the test reads it back to make assertions:

1. init creates it — writes the MigrationConfig (topics, cluster link, gateway CRs, PauseConsumerOffsetSync: true, initial FSM state).
2. The test then calls readMigrationState(t, cfg, stateFile) (which kubectl exec cats the file out of the pod) to pull the generated migrationID to pass to execute.
3. execute reads it to know where to resume, drives the FSM forward, and rewrites it after each transition.
4. After execute, the test reads it again to assert CurrentState == "switched" and PauseConsumerOffsetSyncFlipped == false (i.e. the drain→disable→restore cycle completed and the state persisted correctly).

Why the name is scenario-specific

Each e2e scenario uses its own state file (…-pause-sync-drain.json, …-pause-sync-happy.json, etc.). Because every scenario also has its own dedicated cluster link and gateway CR, giving each its own state file keeps the tests fully independent — one test's persisted FSM state can't bleed into another's, so they can run in any order (or in parallel) without collisions.

So in short: it's the durable handoff between init and execute, and the test's window into the FSM's final state — scoped per-scenario for isolation.

adrian-januzi: Ah okay part of the `setup.sh` script - I was unaware, thanks.

adrian-januzi (approved):
lgtm

PR #385, opened 2026-07-10
fix(migration): drain old gateway pods before unrouted-producer detection

bmartin494:
## Problem

`kcp migration execute --detect-unrouted-producers-duration <N>` can report a rogue producer and abort a migration (`abort_fence` → unfence → rollback to `initialized`) even when **zero** rogue producers exist — only a single legitimate producer sending through the gateway.

**Root cause:** on fence, Kubernetes does a rolling pod restart. `FenceGateway` returned as soon as the new fenced pod reported Ready, but k8s keeps the **old, still-unfenced pod** serving live traffic for ~30–40s (readiness-probe `initialDelaySeconds`) before scaling it down. A legitimate gateway-routed producer keeps landing writes on the source cluster through that old pod, so the detector's two-snapshot source-offset comparison sees offsets climbing and can't distinguish it from a real rogue producer.

## Fix

When detection is enabled, `FenceGateway`:
1. captures the gateway pod UIDs **before** applying the fenced CR;
2. waits until the operator has **observed** the fenced CR (`status.observedGeneration >= metadata.generation`) — so a slow operator reconcile can't let the rollout-detection window expire against a still-unfenced pod;
3. waits until those **old pods are actually gone** (not just until the new pod is Ready).

So the detector's first snapshot lands on a genuinely quiescent source. Reuses `GetGatewayPodUIDs` + `WaitForGatewayPods` (previously dead code). Gated on `--detect-unrouted-producers-duration > 0`, so the default fence path is unchanged.

Two latent bugs in `WaitForGatewayPods`, surfaced by wiring it into a live path, are also fixed:
- `timeout == 0` now means **no deadline** (matching `WaitForGatewayReady`), instead of timing out immediately.
- completion gates on `oldPodsRemaining == 0 && deploymentRolloutComplete` rather than the pod count captured before the patch. If capture races an in-flight rollout (a surge leaves 2 pods where desired is 1), the captured count is inflated and `newPodsReady` can never reach it — a deadlock. The Deployment's own desired/ready counts are immune to that race.

## Validation (e2e, fresh cluster)

- ✅ `TestMigrationE2E_RogueProducerFalsePositive` — passes; fence waits out reconcile + old-pod drain, detection passes, migration reaches `switched`.
- ✅ `TestMigrationE2E_RogueProducerDetection` — real direct-to-source rogue still caught (Phase A abort + gateway restore), and the resume path (Phase B, surge-capture) now completes to `switched` instead of deadlocking.

Plus unit tests for `WaitForGatewayPods` (incl. the surge-capture completion) and `WaitForGatewayObservedGeneration`.

## Independent review

Reviewed for security, correctness, and design. No critical/high findings; no confirmed bugs. Safety property confirmed: the check cannot be silently bypassed to let a real rogue through — every degraded path fails toward a spurious abort (false positive), never a false negative. The one consensus concern (the Phase-1 "no rollout detected" gap under a slow operator) is addressed by the reconcile-wait above.

### Tracked follow-ups (not blocking)
- **No-deadline hang:** with `--rollout-timeout` unset (0), a pod stuck `Terminating` (finalizer/PDB/NotReady node) can hang the fence indefinitely with clients fenced; consider a sane default cap and/or `signal.NotifyContext`.
- **Pod selector:** `GetGatewayPodUIDs` assumes `app=<gatewayName>`; if CFK labels ever differ it returns empty and the old-pod guarantee silently degrades. Prefer deriving the selector from the Deployment's `spec.selector`.
- **Transient-error tolerance:** a single k8s API error aborts the multi-minute drain; offset sweeps tolerate 3.
- **`initialPodCount == 0`** (zero pods pre-fence): warn + fall back.
- **Asymmetry:** `unfenceGateway`/`SwitchGateway` still use the weaker readiness wait (fine — no detection follows them — but the rollback's "traffic restored" message can slightly over-claim).


bmartin494:
### Merged latest `main` in (`c9c43f72`) — heads-up on the CI e2e capacity risk

Brought the branch up to date with `main`. Conflicts were trivial (scenario-list unions: `pause-sync-drain` from #381 + `rogue-producer-false-positive`). `go build ./...` clean; full unit suite green (55 packages).

**The migration e2e (`make test-migration`) may go red in CI for a capacity reason, not a code defect — flagging so it isn't misread:**

- This branch now provisions **10** migration scenarios (main has 9). Each scenario runs a gateway pod, and a fence does a rolling update that briefly **surges a second pod**.
- `setup.sh` starts minikube with `--memory=8192` (8 GB) in both local and CI.
- Reproduced locally: with 10 scenarios the fence's surge pod fails to schedule (`FailedScheduling: Insufficient memory`), so `deploymentRolloutComplete` never becomes true and the fence wait (no deadline) hangs → the detection-enabled tests (`RogueProducerFalsePositive`, `RogueProducerDetection`) time out.
- Local diagnosis was partly muddied by a 9 GB Docker Desktop VM; CI's `-2` agent has more host RAM, so it **may** fit 10 scenarios where the local VM couldn't — but main-green only proves **9** fit, not 10.

**If the e2e block goes red on `FailedScheduling` / a fence-wait timeout:** the fix is infra, not the change — bump `setup.sh --memory` above 8192 (CI agent has the headroom). Also note this concretely demonstrates the tracked follow-up that a Pending/unschedulable pod hangs the no-deadline fence wait — worth bounding.

The fix itself was e2e-validated green pre-merge (both tests) on a fresh cluster; the merge only added scenario-list unions plus clean auto-merges in `workflow.go`/`gateway.go`.

pinaruyar:
TestMigrationE2E_RogueProducerFalsePositive — PASS (90.06s)

The scenario that directly reproduces your reported bug: one legitimate producer running through the gateway's proxy the whole time, zero rogue producers.

🚧 Fencing gateway...
   ✔ Fenced gateway CR applied
   ↳ Waiting for gateway readiness...
   ↳ Waiting for gateway reconcile...
   ↳ 0/1 new pods ready, 1 old pods remaining   ← old unfenced pod still tracked
   ↳ 0/1 new pods ready, 1 old pods remaining
   ...
   ↳ 1/1 new pods ready, 1 old pods remaining   ← new pod Ready, but old one NOT gone yet
   ↳ 1/1 new pods ready, 1 old pods remaining
   ...
   ↳ 1/1 new pods ready, 0 old pods remaining   ← only now does the wait complete
   ✔ Gateway fenced and ready

🔍 Checking for unrouted producers...
   ↳ Monitoring source offsets for 10s...
   ✔ Source offsets stable — no unrouted producers detected   ← the bug fix in action

📤 Promoting mirror topics...
   ...
✅ Migration complete!

That "1/1 new pods ready, 1 old pods remaining" line, sitting there for several poll ticks even after the new pod is Ready, is exactly the race window the PR closes — the old code would have declared "fenced and ready" the moment it hit "1/1 new pods ready" and started the detection clock right there, while this stale pod was still serving live writes.

TestMigrationE2E_RogueProducerDetection — PASS (220.40s)

Regression check: a real rogue producer must still be caught, and the resume path must not deadlock.

Phase A — producer writing directly to the source bootstrap (bypassing the gateway):
🔍 Checking for unrouted producers...
   ↳ Monitoring source offsets for 10s...
   ⚠️ Unrouted producers detected — removing fence to restore traffic
   ✔ Initial gateway CR applied
   ↳ Waiting for gateway readiness...
   ✔ Gateway unfenced — traffic restored to pre-migration state
ERROR failed to execute migration: ... unrouted producers detected:
  topic e2e-test-topic-rogue-producer partition 0: offset 258 → 289 (+31, ~3 msg/s)
  topic e2e-test-topic-rogue-producer partition 1: offset 239 → 273 (+34, ~3 msg/s)
  topic e2e-test-topic-rogue-producer partition 2: offset 266 → 301 (+35, ~4 msg/s)
Correctly aborted, named the exact topic/partitions and rates, rolled the gateway back, migration state returned to initialized.

Phase B — rogue producer stopped, resume:
↳ 0/2 new pods ready, 2 old pods remaining
...
↳ 1/2 new pods ready, 0 old pods remaining
✔ Gateway fenced and ready
...
✔ Source offsets stable — no unrouted producers detected
📤 Promoting mirror topics...
🔄 Switching gateway to Confluent Cloud...
✅ Migration complete!

internal/services/migration/workflow.go
adrian-januzi, line 1: Before merging, we probably want to show the command as it is currently hidden: https://github.com/confluentinc/kcp/blob/test/rogue-producer-fence-race-repro/cmd/migration/execute/cmd_migration_execute.go#L103

adrian-januzi (commented):
Before merging should we also test this against the new hot reload CFK image to ensure that this doesn't break due to a lack of pod rollout?

This is kind of a wider `kcp migration execute` question too - should we confirm hot reload hasn't broken the rest of the command?

internal/services/migration/workflow.go
bmartin494: removed MarkHidden so --detect-unrouted-producers-duration shows in migration execute --help, and re-advertised it in the getting-started guide (Verify fence phase back in the cutover table). Dropped the [Experimental] tag

adrian-januzi (approved):

PR #390, opened 2026-07-21
Fix/UI topic cleanup policy badges

pinaruyar:
## Description

Fixes a mislabeling bug in the `kcp ui` Topics tab: topics with a combined `cleanup.policy` of `compact,delete` (or `delete,compact`) were shown with only a "Delete" badge, hiding the fact that they're actively compacting. This matters beyond cosmetics — compacted vs. non-compacted partition counts factor directly into Confluent Cloud Enterprise eCKU sizing, so an inaccurate badge risks under-counting compacted partitions during migration sizing conversations. Also clarifies two summary-card labels and a column header that were misleading about what they actually count.

---

## Changes Made

- [x] Fix: per-topic "Type" column now renders **both** Compact and Delete badges independently when a topic's `cleanup.policy` contains both tokens (previously an exact-string-match bug meant `compact,delete` only ever showed "Delete"). Column header renamed from "Type" to "Cleanup Policy" to match.
- [x] Fix: "Total Topics" and "Total Partitions" summary cards renamed to "Total Non-Internal Topics" / "Total Non-Internal Partitions" — verified against the backend (`CalculateTopicSummaryFromDetails` in `internal/types/kafka.go`) that these fields only ever count non-internal (`__`-prefixed excluded) topics/partitions; the old labels implied a grand total that included internal topics, which they don't.
- [x] Fix (unrelated, pre-existing): `internal/types/state_schema_test.go` used the deprecated `reflect.Ptr` alias, which failed the repo's `golangci-lint` pre-commit gate on bare `main` regardless of this change. Fixed with the tool-suggested `reflect.Pointer` equivalent in its own commit so it doesn't block this PR.
- [x] Any breaking changes — none. Purely additive/cosmetic UI changes; no state file schema or API changes.
- [x] Documentation updates — not needed (no user-facing docs describe this UI in that level of detail today).

---

## Testing

- [x] New tests added/updated — new Playwright e2e spec `cmd/ui/frontend/tests/e2e/cluster-topics-cleanup-policy.spec.ts` (5 cases: delete-only, compact-only, `compact,delete`, `delete,compact` reversed-order, and an internal+compacted topic), with a new fixture `state-topic-cleanup-policy.json`.
- [x] Manual testing completed — built the binary and drove the UI directly against the fixture to confirm rendering before relying on the automated spec.
- [x] All tests pass — `yarn type-check` clean; e2e spec 5/5 passing (RED confirmed pre-fix: 2/5 failing on the combined-policy cases, exactly as expected; GREEN post-fix: 5/5).

**Test Instructions:**

1. `cd cmd/ui/frontend && yarn test:e2e tests/e2e/cluster-topics-cleanup-policy.spec.ts`
2. Or manually: `make build && ./kcp ui --state-file cmd/ui/frontend/tests/e2e/fixtures/state-topic-cleanup-policy.json`, open the `cleanup-policy-test-cluster` cluster → Topics tab, and check `audit-log` / `reversed-policy-topic` rows show both Compact and Delete badges.

---

## Checklist

- [x] Code follows project style guidelines (Tailwind inline classes, `@/` aliases, named exports — matches existing `ClusterTopics.tsx` conventions)
- [x] Self-review completed
- [ ] Documentation updated (if needed) — n/a
- [x] No breaking changes (or breaking changes documented) — none

---

## Screenshots/Demo/Output
Screenshot 2026-07-21 at 11 50 26
Screenshot 2026-07-21 at 11 50 54

### DIFF
--- a/cmd/ui/frontend/src/components/explore/clusters/ClusterTopics.tsx
+++ b/cmd/ui/frontend/src/components/explore/clusters/ClusterTopics.tsx
@@ -1,4 +1,4 @@
-import { formatRetentionTime } from '@/lib/utils'
+import { formatRetentionTime, parseCleanupPolicies } from '@/lib/utils'
 import type { KafkaAdminInfo, Topic } from '@/types'
 
 interface ClusterTopicsProps {
@@ -100,15 +100,22 @@ export const ClusterTopics = ({ kafkaAdminInfo }: ClusterTopicsProps) => {
                     {topic.replication_factor}
                   </td>
                   <td className="py-3 text-center">
-                    <span
-                      className={`px-2 py-1 text-xs rounded ${
-                        topic.configurations['cleanup.policy'] === 'compact'
-                          ? 'bg-blue-100 text-blue-800 dark:bg-accent/20 dark:text-accent'
-                          : 'bg-green-100 text-green-800 dark:bg-green-900 dark:text-green-200'
-                      }`}
-                    >
-                      {topic.configurations['cleanup.policy'] === 'compact' ? 'Compact' : 'Delete'}
-                    </span>
+                    <div className="flex items-center justify-center gap-1">
+                      {parseCleanupPolicies(topic.configurations['cleanup.policy']).includes(
+                        'compact'
+                      ) && (
+                        <span className="px-2 py-1 text-xs rounded bg-blue-100 text-blue-800 dark:bg-accent/20 dark:text-accent">
+                          Compact
+                        </span>
+                      )}
+                      {parseCleanupPolicies(topic.configurations['cleanup.policy']).includes(
+                        'delete'
+                      ) && (
+                        <span className="px-2 py-1 text-xs rounded bg-green-100 text-green-800 dark:bg-green-900 dark:text-green-200">
+                          Delete
+                        </span>
+                      )}
+                    </div>
                   </td>
                   <td className="py-3 text-center text-foreground">
                     <div className="relative group">
--- a/cmd/ui/frontend/src/components/explore/clusters/ClusterTopics.tsx
+++ b/cmd/ui/frontend/src/components/explore/clusters/ClusterTopics.tsx
@@ -32,13 +32,13 @@ export const ClusterTopics = ({ kafkaAdminInfo }: ClusterTopicsProps) => {
             <div className="text-2xl font-bold text-foreground">
               {kafkaAdminInfo.topics.summary.topics}
             </div>
-            <div className="text-sm text-muted-foreground">Total Topics</div>
+            <div className="text-sm text-muted-foreground">Total Non-Internal Topics</div>
           </div>
           <div className="bg-secondary rounded-lg p-4 transition-colors">
             <div className="text-2xl font-bold text-foreground">
               {kafkaAdminInfo.topics.summary.total_partitions}
             </div>
-            <div className="text-sm text-muted-foreground">Total Partitions</div>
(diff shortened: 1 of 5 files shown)


pinaruyar:
## Known Issue / Follow-up (not fixed in this PR)

This PR relabels the UI to say "Non-Internal" explicitly, but that's a display-layer patch over a naming ambiguity that actually lives in the backend schema itself and should be fixed properly in a future PR.

**The issue:** `internal/types/kafka.go`'s `TopicSummary` struct has `Topics`/`TotalPartitions` sitting as peer fields next to `InternalTopics`/`TotalInternalPartitions`. The names read like a superset/subset relationship (you'd expect `Topics` to include `InternalTopics`), but they're actually two disjoint categories — `Topics + InternalTopics` = the real total, `Topics` alone is not. Every consumer of this struct (this UI, the `kcp discover` table, the `kcp scan` healthcheck report) inherited this ambiguity; the healthcheck report already worked around it by labeling its bullet "User topics" instead of "Total topics," which is a better precedent than what the UI had.

**Risks of fixing it properly:**
- Renaming the JSON tags (e.g. `topics` → `non_internal_topics`, `total_partitions` → `total_non_internal_partitions`) is a breaking change to the versioned `kcp-state.json` wire format, not just a code rename — every stored/exported state file in the wild has the old key names.
- Requires a `CurrentSchemaVersion` bump (currently `1`, in `internal/state/migrate/migrate.go`) plus a new upcaster in `internal/state/migrate/steps.go`. Every existing upcaster there only reshapes *top-level* keys; this rename lives nested inside `topics.summary`, under `msk_sources.regions[].clusters[]` **and** `osk_sources.clusters[]`, so the new step needs to recurse into arrays — more complex than anything currently in that file.
- The `step.appliesWhen(era, buildVersion)` signature doesn't currently take `schemaVersion` at all (only two root-shape "eras," B/C, have ever been distinguished that way) — wiring a v1→v2 rename means extending that mechanism, not just registering a new step.
- `internal/types/state_schema_test.go` walks the struct via reflection and diffs against `internal/types/testdata/state_schema.golden` — any field rename fails that test until the golden file is regenerated.
- Every non-UI consumer needs updating in lockstep: `cmd/discover/discoverer.go` (table column), `cmd/healthcheck/render.go` (bullet list), plus the frontend TS type (`cmd/ui/frontend/src/types/aws/msk.ts`) and this component's accessors.

**Steps to fix it (future PR):**
1. Rename the Go struct fields + JSON tags in `internal/types/kafka.go` (`Topics`→`NonInternalTopics`, `TotalPartitions`→`TotalNonInternalPartitions`; `InternalTopics`/`TotalInternalPartitions` are already correctly named, no change needed).
2. Update all consumers: `cmd/discover/discoverer.go`, `cmd/healthcheck/render.go`, `cmd/ui/frontend/src/types/aws/msk.ts`, `ClusterTopics.tsx` accessors.
3. Bump `CurrentSchemaVersion` to `2` in `internal/state/migrate/migrate.go`.
4. Extend the `step` mechanism in `internal/state/migrate/steps.go` to gate on `schemaVersion` (not just era), and add the v1→v2 upcaster that recurses into `msk_sources.regions[].clusters[]` and `osk_sources.clusters[]` to rename the keys inside each `topics.summary` object.
5. Regenerate `internal/types/testdata/state_schema.golden` (flip `updateSchemaGolden` locally per the comment in `state_schema_test.go`).
6. Verify `kcp state upgrade` round-trips an old (`schema_version: 1`) real-world file correctly through the new upcaster before shipping.

adrian-januzi (approved):
lgtm

PR #393, opened 2026-07-22
fix(ui): rename Compact Topics card to Compact Non-Internal Topics

pinaruyar:
kafkaAdminInfo.topics.summary.compact_topics only counts non-internal compacted topics (verified against CalculateTopicSummaryFromDetails in internal/types/kafka.go and against a real customer state file: 315 non-internal + 35 internal = 350 true total, but the field/card only ever reflected the 315). Same scope issue the prior PR (#390) fixed for Total Topics/Total Partitions, applied here for consistency.

## Description

Topics Overview card: 
  - Split into Non-Internal / Internal tabs instead of a flat row of cards — each tab shows 4 stats (Topics,
  Compact Topics, Partitions, Compact Partitions) scoped to that type, so all 8 real summary fields are now
  visible without a wall of near-duplicate labels.
  - Dropped the redundant "Total" prefix on the Non-Internal labels for consistency with the Internal ones.
  
  All Topics table:
  - Moved the "Internal" badge out of the Topic Name cell into its own Type column (matching the existing
  Cleanup Policy column), and added a "Non-Internal" badge that didn't exist before — previously non-internal
  topics got no badge at all.
  - Added a filter icon next to "All Topics" with two checkbox groups — Type (Non-Internal/Internal) and
  Cleanup Policy (Compact/Delete) — all four independently toggleable and additive (both checked in a group
  shows everything; a topic matches on cleanup policy if it has any checked policy, since a topic can carry
  both compact and delete at once).

---

## Changes Made

- [x] Feature/bug fix/improvement description
- [ ] Any breaking changes
- [ ] Documentation updates

---

## Testing

- [ ] New tests added/updated
- [ ] Manual testing completed
- [ ] All tests pass

**Test Instructions:**

---

## Checklist

- [ ] Code follows project style guidelines
- [ ] Self-review completed
- [ ] Documentation updated (if needed)
- [ ] No breaking changes (or breaking changes documented)

---

### DIFF
--- a/cmd/ui/frontend/package.json
+++ b/cmd/ui/frontend/package.json
@@ -51,7 +51,7 @@
   "resolutions": {
     "lodash": "^4.18.0",
     "lodash-es": "^4.18.0",
-    "fast-uri": "^3.1.2"
+    "fast-uri": "^3.1.4"
   },
   "devDependencies": {
     "@eslint/js": "10.0.1",
--- a/cmd/ui/frontend/src/components/explore/clusters/ClusterTopics.tsx
+++ b/cmd/ui/frontend/src/components/explore/clusters/ClusterTopics.tsx
@@ -1,3 +1,5 @@
+import { useState } from 'react'
+import { Tabs } from '@/components/common/Tabs'
 import { formatRetentionTime, parseCleanupPolicies } from '@/lib/utils'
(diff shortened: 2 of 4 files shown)


adrian-januzi (approved):

adrian-januzi (approved):
lgtm

PR #397, opened 2026-07-27
fix(migration): verify the operator accepted the gateway CR before reporting success

bmartin494:
## Description

`kcp migration execute` could report a **successful migration while the gateway cutover never happened**.

Found on 2026-07-27 while setting up the new end-to-end live-cluster test infrastructure (real CFK operator, real Confluent Cloud clusters). kcp printed `No pod restart required` then `✅ Migration complete!` — but the gateway was still fenced, blocking all client traffic. The CFK operator had rejected the switchover spec:

```
generation=4, observedGeneration=3
conditions: {"message":"secretRef kcp-perf-plain-jaas not found",
             "reason":"ApplyFailed","status":"False",
             "type":"platform.confluent.io/cluster-ready"}
```

The trigger was a genuinely missing k8s secret referenced by the switchover CR — a real misconfiguration in the test environment being stood up. The bug is that **kcp reported success anyway**. No customer migration was affected, but the same code path runs in one.

### Root cause

`SwitchGateway` applied the CR then called `WaitForGatewayReady`, which only inspects the **apps/v1 Deployment** (`observedGeneration`, `updatedReplicas`, `availableReplicas`). When the operator refuses a Gateway CR it never touches the Deployment, so the Deployment sat complete and healthy — running the *previous* generation's fenced pods. The readiness wait's 10s detection window saw a steady, complete Deployment, concluded "no pod restart was required", and returned `nil`.

No Deployment field can distinguish *"no spec change needed"* from *"the operator refused the spec"*. Meanwhile the operator's own diagnosis was sitting in the Gateway CR's `status.conditions`, which nothing in the codebase read.

### Broader than switchover

The fence path already guards against this with `WaitForGatewayObservedGeneration`, whose doc comment describes this exact failure mode — but it was called at **exactly one site**, inside `if detecting` in `FenceGateway`. Every path relying solely on `WaitForGatewayReady` had the same silent-failure hole:

| Path | Before |
| --- | --- |
| `SwitchGateway` | ❌ reports a completed migration for a switchover that never happened |
| `FenceGateway`, `--detect-unrouted-producers` **off** (the default) | ❌ proceeds to promote topics with the gateway never fenced |
| `FenceGateway`, detection **on** | ⚠️ guarded, but hangs forever on rejection (see below) |
| `unfenceGateway` | ❌ reports traffic restored during rollback while still fenced |

### Why mirroring the existing guard wasn't enough

Adding only `WaitForGatewayObservedGeneration` would **not** have produced a clean error. On operator rejection `observedGeneration` never advances, so with the default `--rollout-timeout 0` (no deadline) kcp would have hung indefinitely. Better than a false success, but poor UX — and that latent hang existed on the fence path already. So the fix also reads the CR's status conditions and fails fast with the operator's own message.

---

## Changes Made

- [x] Replace `WaitForGatewayObservedGeneration` with **`WaitForGatewayAccepted`** (renamed — the old name would be a lie now that it can return a rejection), called after every apply and before the Deployment-based waits, at all four sites: `SwitchGateway`, both `FenceGateway` branches, and `unfenceGateway`.
- [x] Each poll races two outcomes:
  - **accepted** — `observedGeneration >= generation` → `nil`. Checked *first*, so a lingering stale condition on an already-accepted spec can't abort a good migration. A no-op re-apply (resume of an already-fenced gateway) still returns immediately.
  - **rejected** — a `status.conditions` entry with `status: False` and a fatal reason → typed **`*GatewayRejectedError`** carrying the operator's own reason and message, so callers can `errors.As` it rather than string-match.
- [x] Guard against false failures, since aborting a healthy migration is its own kind of outage:
  - a failing condition must persist for a **30s settle window** before it aborts;
  - the window measures **one unbroken run** of failures and resets the moment the condition clears, so an operator flapping while it converges can't accrue its way to a spurious abort;
  - only reasons that are **catalogued** (`ApplyFailed`, the one actually observed against a real CFK operator) or follow CFK's **`<Verb>Failed`** convention qualify. Unknown reasons are treated as mid-reconcile noise and waited on.
- [x] When `--rollout-timeout` is set *shorter* than the settle window, a pending rejection is returned instead of a bare timeout — strictly more informative and still accurate.
- [x] Shared helper `MigrationActions.waitForGatewayAccepted(ctx, config, step)` so fence, switchover and unfence cannot drift apart; `step` names the phase in the error.
- [ ] Any breaking changes — none. `--rollout-timeout` semantics are unchanged; the interface method rename is internal (`gateway.Service` has no external implementors).
- [ ] Documentation updates — none needed; rationale lives in the doc comments on `WaitForGatewayAccepted` and `waitForGatewayAccepted`.

### Operator error now surfaced

```
gateway "kcp-perf" was rejected by the Confluent operator: secretRef kcp-perf-plain-jaas not
found (reason: ApplyFailed, condition: platform.confluent.io/cluster-ready, generation: 4,
observedGeneration: 3)
```

plus a remediation hint with what the error can't carry:

```
⚠️ Confluent operator rejected the switchover gateway spec. Inspect its view of the gateway:
   kubectl -n <ns> get gateway <name> -o jsonpath='{.status.conditions}'
```

---

## Testing

- [x] New tests added/updated — 20 new/renamed
- [x] All tests pass — **170 passing, 0 failing** across `internal/services/gateway` and `internal/services/migration`; full `go test ./...` green; `-race -count=3` clean (the timing-based tests are not flaky)
- [ ] Manual testing completed — the live-cluster e2e run that surfaced this is the manual repro; a deliberate re-run against a missing secret is worth doing before release

The core regression test reproduces the observed CR exactly — `generation=4`, `observedGeneration=3`, `ApplyFailed`, the real secret name — and asserts `WaitForGatewayReady` is **never reached**, pinning the specific fall-through that produced "No pod restart required".

```
--- PASS: TestWorkflow_SwitchGateway_OperatorRejection_FailsWithOperatorMessage (0.00s)
--- PASS: TestWaitForGatewayAccepted_Rejected_ReturnsTypedErrorWithOperatorMessage (0.03s)
--- PASS: TestWaitForGatewayAccepted_Rejected_HonoursSettleWindow (0.10s)
--- PASS: TestWaitForGatewayAccepted_TransientFailureThenAccepted_ReturnsNil (0.05s)
--- PASS: TestWaitForGatewayAccepted_ClearedConditionResetsSettleTimer (0.17s)
--- PASS: TestWaitForGatewayAccepted_UnknownReason_KeepsWaiting (0.08s)
--- PASS: TestWaitForGatewayAccepted_TrueConditionWithFailedReason_NotARejection (0.06s)
--- PASS: TestWaitForGatewayAccepted_TimeoutShorterThanSettleWindow_ReportsRejection (0.06s)
--- PASS: TestWaitForGatewayAccepted_AcceptedDespiteFailedCondition_ReturnsNil (0.00s)
--- PASS: TestWorkflow_FenceGateway_OperatorRejection_DoesNotProceed (0.00s)
--- PASS: TestWorkflow_SwitchGateway_WaitsForAcceptanceBeforeReadiness (0.00s)
--- PASS: TestWorkflow_SwitchGateway_NonRejectionWaitError_IsWrapped (0.00s)
--- PASS: TestWorkflow_SwitchGateway_PassesRolloutTimeoutToAcceptanceWait (0.00s)
--- PASS: TestWorkflow_UnfenceGateway_OperatorRejection_Fails (0.00s)
--- PASS: TestWorkflow_UnfenceGateway_WaitsForAcceptanceBeforeReadiness (0.00s)
--- PASS: TestIsFatalGatewayConditionReason (0.00s)
```

One existing test changed premise deliberately: `TestWorkflow_FenceGateway_DetectionDisabled_UsesReadyWaitNotUIDDiffing` previously asserted the reconcile wait was **not** called with detection off. It now asserts the opposite — the acceptance wait runs on every path — while still asserting the fence keeps its lightweight readiness-only wait and never touches pod UIDs.

**Test Instructions:**

```bash
make build-frontend   # required before any Go test (go:embed)
go test ./internal/services/gateway/ ./internal/services/migration/ -v
go test -race -count=3 ./internal/services/gateway/ ./internal/services/migration/
```

To exercise it against real infra: point a switchover CR at a `secretRef` that doesn't exist and run `kcp migration execute`. Before this change it reports `✅ Migration complete!`; after, it fails within ~30s with the operator's message.

---

## Checklist

- [x] Code follows project style guidelines (emoji standard respected: `⚠️` on `slog.Warn`, `❌` on `slog.Error`, none in `fmt.Errorf`)
- [x] Self-review completed
- [x] Documentation updated (if needed) — doc comments only
- [x] No breaking changes

---

## Second commit: the e2e rig had the same bug

The migration e2e suite went red on the first commit — correctly. Investigating it turned up a **pre-existing gap in the test fixtures**, identical to the bug this PR fixes.

`gateway-switchover.yaml` references four secrets that `setup.sh` never created: `destination-kafka-tls`, `plain-jaas`, `file-store-config`, `file-store-noauth-credentials`. (The initial/fenced CRs need none — they reach `source-kafka` over PLAINTEXT with `auth: passthrough`; only switchover talks SASL_SSL to `destination-kafka`.) So CFK refused every switchover spec with `cluster-ready=False reason=ApplyFailed "secretRef destination-kafka-tls not found"`. `integration-tests/` was byte-identical to `main`, so this predates the PR.

**And the suite could not have caught it.** Server-side apply succeeds and bumps `metadata.generation` even when the operator then refuses to reconcile, so `verify_gateway_cr`'s spec assertions passed; the backing Deployment stayed healthy on the previous generation; and nothing read `status.observedGeneration`. **Every e2e switchover has been a silent no-op** — the exact failure mode of this PR, sitting in the rig meant to exercise it.

Fixed by creating the four secrets, with shapes from `docs/assets/gateway-switchover/switchover-none-to-saslplain/`. For trust material, reuse the truststore CFK already generates for `destination-kafka`'s `tls.autoGeneratedCerts` rather than building a parallel one — it already trusts the `ca-pair-sslcerts` CA that signed the cert being verified, so it cannot drift from the CA in use, and it needs no `keytool`/JDK on CI agents. `setup.sh` hard-fails with the lookup command if that secret is ever absent, rather than producing a gateway that cannot connect.

Plus **`assertGatewayAccepted`** — asserts `status.observedGeneration` caught up to `metadata.generation`, dumping conditions on failure. This is the assertion whose absence let the no-op hide, and it is independent of kcp's own acceptance wait.

### Verification on real infrastructure

Validated on two independently created minikube/CFK clusters before pushing:

| Check | Result |
|---|---|
| CFK generated-secret name | ✅ both cold clusters |
| Truststore tracks each cluster's own freshly generated CA | ✅ fingerprints differ per cluster, match exactly |
| `openssl verify` broker cert against that truststore | ✅ |
| Cert SAN matches the CR endpoint | ✅ `destination-kafka.confluent.svc.cluster.local` |
| Operator **accepts** the switchover | ✅ `observedGeneration == generation`, no `ApplyFailed` |
| `TestMigrationE2E` (baseline) | ✅ green, 10/10 subtests, 88s |

```
--- PASS: TestMigrationE2E (88.18s)
    --- PASS: TestMigrationE2E/execute (79.61s)
    --- PASS: TestMigrationE2E/verify_gateway_cr (0.00s)      <- includes assertGatewayAccepted
    --- PASS: TestMigrationE2E/verify_mirror_topics_stopped (0.08s)
    --- PASS: TestMigrationE2E/verify_pod_rollout (0.00s)
ok  	github.com/confluentinc/kcp/integration-tests/migration	89.204s
```

The switchover performed a real pod rollout (`1/2 pods ready` → `✔ Gateway switchover complete`) — the first genuine cutover this suite has ever done. Runtime is unchanged (88s vs 91s before).

**Scope of that claim:** I validated the `baseline` scenario only. All ten scenarios render from the same three templates with only names substituted by `sed`, and the four secrets are cluster-wide rather than per-scenario, so the rejection cause is fixed for all of them — but that is an inference from fixture structure, not an observation. CI exercises all ten.

### Note for anyone running this suite locally

The full suite **cannot pass on a 9GB Docker Desktop VM.** Each of the ten scenario gateways reserves 512Mi through an *init container* request — invisible in `kubectl get pods -o custom-columns` on `.spec.containers` — totalling ~8.6Gi of the node's ~9.2Gi allocatable. A fence needs a further 512Mi surge pod for its rolling update, which cannot schedule, and the run stalls on `Insufficient memory` until the test times out. CI agents have the headroom; `setup.sh`'s `--memory=8192` is deliberately left alone. To run locally, either raise the Docker Desktop memory limit or delete the scenario gateways you are not exercising.

Also: **killing `go test` does not kill the `kcp` process inside the runner pod.** `kubectl exec` does not propagate the kill, so an orphaned `kcp migration execute` keeps running, and can resume and complete the migration later — consuming the scenario (promotion is terminal, so `init` then refuses with `mirror topics are not active ... (status: STOPPED)`).

---

## Relationship to #398

[#398](https://github.com/confluentinc/kcp/pull/398) ("actually validate the gateway CRs instead of claiming to") landed on main while this was open, and implements the pre-flight `secretRef` existence check this PR's description previously listed as a follow-up. **That follow-up note is now obsolete — it is done.**

The two are complementary layers, not duplicates:

| | When it runs | What it catches |
|---|---|---|
| **#398** `ValidateGatewayCRs` | at `init`, before any traffic is touched | a referenced secret that does not exist — the cheapest possible failure, no fence, no promotion |
| **This PR** `WaitForGatewayAccepted` | after each apply, before the Deployment waits | *any* operator rejection, including what no pre-flight check can foresee: an invalid spec, an admission webhook, a quota, or a secret deleted between validation and apply |

Neither subsumes the other. #398 makes the common case fail early and cheaply; this makes a rejection impossible to mistake for success no matter the cause, on all four apply sites (fence ×2, switchover, unfence).

`origin/main` is merged in and the one conflict (`mock_test.go` — #398's new `ValidateGatewayCRs` signature vs this branch's `waitForGatewayAcceptedFn` rename) is resolved by taking both. Post-merge: full `go test ./...` green, `-race -count=2` clean on both changed packages, `golangci-lint` 0 issues.

Note also that main does **not** create the four switchover secrets the e2e rig references. Since #398 makes a missing `secretRef` fatal at `init`, the second commit here is what unblocks that suite.

---

## Follow-up, not in this PR

Nothing outstanding. The pre-flight validation that was listed here has landed separately as #398 — see the section above.


adrian-januzi (approved):

adrian-januzi (changes requested):
merge conflicts

adrian-januzi (approved):

PR #399, opened 2026-07-28
feat: add kcp migration txn-discovery

adrian-januzi:
## Summary

Adds `kcp migration txn-discovery`, a command that observes a live Kafka cluster for a set window, reconstructs which topics are coupled by transactions, and writes a reviewable `txn-discovery.yaml` alongside a live JSONL audit trace explaining every grouping decision.

Topics coupled by a Kafka transaction must migrate together. Move one without the others and an in-flight transaction spans two clusters, which no cluster link can reconcile. Until now a migration engineer had no way to discover that coupling from kcp — they inferred it from application knowledge, which is exactly the knowledge a customer engagement lacks.

This is a sarama-based reimplementation of the franz-go POC in `utils/transaction-discovery/`, with real test coverage and kcp's full authentication surface (including MSK IAM, which the POC lacks). The POC is untouched and still builds independently.

## Why the Kafka layer was rebuilt rather than ported

`sarama.ConsumerMessage` discards the record-batch header, so the high-level consumer cannot see the producer ID that ties a transactional offset commit to its transaction. That producer ID is what recovers consumed input topics for **non-Streams** exactly-once applications, where no naming convention exists to correlate on. Keeping the capability meant hand-rolling a fetch loop over raw `sarama.Broker.Fetch`, which brings partition discovery, leader resolution, offset tracking, failover, and — most easily missed — **aborted-transaction filtering** into this codebase's ownership. `Isolation: ReadCommitted` does not filter aborted records at the wire level; it bounds the fetch at the last stable offset and attaches an `AbortedTransactions` list, and discarding those records is client-side work.

## What landed

14 planned units, executed test-first:

| Area | Detail |
|---|---|
| Decoders | `TransactionLogKey`/`TransactionLogValue` and a hand-rolled `__consumer_offsets` commit-key decoder, with literal-byte fixtures rather than an encoder from the library under test |
| Tail component | Continuous resumable reader over raw `Broker.Fetch`: abort filtering, partial-batch handling, per-partition liveness, LSO-based lag, Fetch-version clamping against the broker's advertised ceiling |
| Sources | `__transaction_state` reader (source of truth), consumer-group enrichment via the Streams naming convention, `__consumer_offsets` tail with exact producer-ID correlation |
| Grouping | Union-find with internal topics filtered **before** the union, deterministic output |
| Output | Counts-only terminal summary, `txn-discovery.yaml`, `--stats-out` JSON, and a change-only JSONL audit log — all three written `0600` |
| Command | Migration's auth flag idiom, fail-fast preflight distinguishing an unknown topic from an authorization failure, `--dry-run` |
| Suites | Single-node docker-compose suite (in CI) and an opt-in three-broker leader-failover suite (not in CI) |

Also: `Metadata.AllowAutoTopicCreation = false` in kcp's shared sarama client construction. sarama defaults it to `true`, so a metadata request naming a nonexistent topic could make a permissive broker create it — turning a discovery tool into one that mutates the cluster it inspects.

## Defects this work surfaced and fixed

Several were latent in the design or the POC rather than introduced here. All were found by tests written to fail first.

- **Integer overflow in the decoder's bounds guard (HIGH).** `need()` compared `pos+n > len(b)`; with `n` from a uvarint, `pos+n` overflowed int64 and went negative, so the guard passed and the slice panicked. There is no `recover()` in kcp and this runs on a reader goroutine, so a ~30-byte malformed record killed a multi-hour observation window. Reachable via two independent paths.
- **Allocation amplification.** A 20-byte record could allocate 80 MB — and still returned a clean error, so an error-only assertion passed while proving nothing. Now bounded by a capped prealloc plus a hard entry cap.
- **Silent buffer over-read.** Stripping the bounds guard from `str()` produced no panic — a record key is a sub-slice of a larger fetch buffer, so reading past `len` into `cap` is legal Go. It spliced 12 bytes of adjacent buffer into a group name. Only the `__consumer_offsets` v2 shape exposed it; v0 and v1 masked it via a trailing bounds-checked read.
- **JSONL forgery.** A newline inside a topic name, without proper encoding, produced a complete well-formed *forged* audit record — in the artifact whose sole purpose is explaining why topics were grouped.
- **`goccy/go-yaml` silently drops TAB characters** from plain scalars: `"a\tb"` round-trips as `"ab"`. Kafka constrains topic names, but `transactional.id` is free-form, so a silently altered id would point an operator at a transaction that does not exist. (`gopkg.in/yaml.v3` was evaluated and is worse — it emits an unparseable document for a leading newline.)
- **Transport errors never reconnected (broke R11).** A broker restart surfaces as a broken pipe, classified as *transport*, and only the *leadership* branch refreshed metadata — so the stale `*sarama.Broker` was never replaced and the loop retried a dead socket. Measured: 98 of 100 partitions read zero records after a restart. Refreshing alone was insufficient: `client.updateBroker` retains the existing broker when the address is unchanged, and sarama never closes a poisoned socket, so the connection is now explicitly discarded.
- **The health line reported OK while 98 partitions were dead.** LSO-based lag is zero for a partition that never fetched, and a loop stuck retrying still counts as running. A per-partition stalled flag now discriminates: an idle partition's empty fetches still *succeed*, so it never latches.
- **Windows: the audit log failed unconditionally.** Go synthesizes `0666` for any writable file, so the `0600` assertion rejected every freshly created file on first run. kcp ships a Windows binary and the audit log is on by default.
- **The POC's `resolveWith` had a silent data-loss path**: it deleted the buffered entry before the send, so a pass racing the end of the window lost the recovery with no counter moving and nothing left for the final flush.
- **Consumer-group enrichment computed correlations and threw them away.** Its admin calls shared one TCP connection with the tail's 100 fetch loops; sarama pipelines per connection and delivers responses in order, so admin requests queued behind long-poll fetches — 6.5–8.1s for `ListConsumerGroups`, 17.6–29.3s for `ListConsumerGroupOffsets`, against 206–461ms on an independent connection. A `5s` interval yielded 4–8 passes per 60s window instead of 13, and the pass that could finally correlate straddled the window's end, where the send took the cancelled branch. The POC never had this — it uses three separate clients. Fixed by giving the enricher its own client (passes per window: **5, 4 → 13, 13**) plus a final pass at window close.
- **Shutdown could hang unboundedly with no escape.** `FinalEnrich`'s 10s context bounded nothing, because `sarama.ClusterAdmin`'s methods take no context — it made `1 + G` blocking round trips after cancellation. Worse, `signal.NotifyContext` is one-shot, so a second Ctrl-C during the drain did nothing; the only exit was `SIGKILL`, which loses all three artifacts. Both halves fixed.

## Security posture

The broker is treated as the untrusted input source. Three broker-internal binary wire formats are parsed, none part of the stable client protocol; malformed, truncated, or format-drifted input must produce a counted decode error, never a panic or an unbounded allocation. Three fuzz targets (`FuzzDecodeValue`, `FuzzDecodeOffsetKey`, `FuzzDecodeKey`) assert no-panic **and** allocation bounded by input size, seeded with the two PoC inputs above so both regressions execute under plain `go test` in CI without `-fuzz`.

Topic names and transactional ids identify customer business structure. The terminal reports counts only; the artifacts carry the names and are all `0600`. **No topic name or transactional id reaches any `slog` call at any level** — `kcp.log` is unconditionally Debug+ and is the file operators attach to support tickets. Credentials are never persisted to an artifact or a log line.

Two security review passes were run in fresh contexts (full diff, then a scoped delta). **Zero Critical, zero High outstanding.** One reviewer finding was rebutted with evidence rather than implemented: the claim that a control marker registers `"\x00"` as a real transactional id is unreachable, because `handle()` returns on the value-decode error before `catalog.Observe`, and `Observe` refuses an empty id. A second was withdrawn by the reviewer on re-test.

## Deliberately deferred

Recorded rather than fixed, with reasoning:

- **Unbounded broker-keyed maps.** `TxnCatalog`, `Accumulator.byTxn`, and the offsets tail's `recovered`/`correlations` grow with distinct broker-supplied strings. Capping them may be wrong for legitimately large clusters, so the decision is left explicit rather than guessed. The pending-sighting buffer *is* bounded, by count and TTL.
- **Hardlink TOCTOU on the audit path.** `O_NOFOLLOW` plus an `Fstat` on the handle closes the symlink race; a hardlink to an existing 0600 file is not covered. `fs.protected_hardlinks=1` and the mode gate narrow it substantially.
- **Decoder ratio.** The worst allocation ratio remains ~74x (a 40-byte struct from ~3 wire bytes, plus append regrowth), now bounded absolutely at ~21.5 MiB per malformed record. Assessed as acceptable: transient, single-threaded, and loudly visible as a decode-error rate.
- **`--insecure-skip-tls-verify` is honoured only by SASL/SCRAM.** The other three TLS-bearing constructors in shared `internal/client` code drop it. The flag is now **rejected** in `PreRunE` when it cannot be honoured, rather than silently ignored; threading it through shared plumbing used by other commands is a separate change.
- **PLAIN-over-TLS is not offered.** `--use-sasl-plain` resolves to kcp's no-TLS variant. The pre-existing cleartext warning does fire; the flag help now says so too, and a warning naming the flag fires once per run.
- **Shutdown latency is bounded only for the enrichment pass.** With a fully unresponsive broker, the drain is dominated by the tail's 100 partition loops, which remain unbounded for the same root cause (blocking sarama calls). Time-to-exit was unchanged at 149s in that scenario; the restored SIGINT abort is what rescues the operator.
- **`--interval` semantics.** Now that the enricher has its own connection, full cadence is restored; the counter added to `--stats-out` makes any future regression visible rather than silent.

## Verification

```
make test-go                 63 packages, 2095 tests, 0 failures
-race (txndiscovery + cmd)   11 packages green
golangci-lint (untagged)             0 issues
golangci-lint (e2e_txndiscovery)     0 issues
golangci-lint (e2e_txndiscovery_ha)  0 issues
make test-txn-discovery      14/14, three consecutive green runs (~231s)
make test-txn-discovery-ha    8/8, ~309s
make test-schema-registry    green
make test-osk-scan           green (all auth methods)
make test-migration          64 subtests, full CFK lifecycle, green
```

The broker-touching suites matter beyond this feature: they are the evidence that disabling auto topic creation is inert for existing commands, which the plan flagged as an assumption. `make test-migration` in particular exercises consumer groups and cluster links — the coordinator-resolution path where auto-create could plausibly have been load-bearing.

`utils/transaction-discovery/` is byte-for-byte unchanged and still builds independently (R26). No `go.mod`/`go.sum` change, and no `kcp-state.json` schema change, so no version bump, upcaster, fixture, or freeze entry (R28).

### The failover suite's negative control

Green alone would not have proved much, so the recovery handler was stubbed. The run **exited 0**, wrote artifacts, and the premise test passed. The offset-continuity assertion **passed** (a stalled reader has no gap — it just stops) and the LSO-lag assertion **passed** (`24/24 live, 0 records of lag` while eight partitions never read again). Only the phase ledger caught it:

```
transactions are missing from the output, by phase:
  map[mid:[zzha-mid-c0-txn zzha-mid-c3-txn zzha-mid-c6-txn]
      post:[zzha-post-c0-txn zzha-post-c3-txn zzha-post-c6-txn]]
  killed node 1, which led __transaction_state partitions [1 3 6 9]
```

Every transactional id in all three phases is pinned to the coordinator being killed — a transactional id's coordinator *is* the leader of the `__transaction_state` partition its footprint lands in. Unpinned, a third of the transactions would sit on undisturbed partitions and a reader that recovered on none could still pass.

## Scope boundaries

Nothing in kcp consumes `txn-discovery.yaml` — wiring these groups into migration batching is the eventual goal and explicitly not this work. Confluent Cloud and MSK Serverless remain unsupported (they do not expose `__transaction_state`; the command fails fast rather than degrading). The compaction blind spot is not solved: a workload whose footprint records were compacted before the window opened is unrecoverable, which is what the keep-up signal exists to tell an operator.

## Note for reviewers

Code review was **deliberately deferred** on this branch rather than run inline — a review conducted inside the session that wrote 19k lines is weaker than the same review run fresh. A `mega-review` pass on this PR in a clean context is the review.

Worth a close look: the tail component (`internal/services/txndiscovery/tail/`) is where the risk concentrates, because it replaces machinery sarama's high-level consumer provides for free and its failure modes are quiet rather than loud. Its coupling to sarama internals — closing brokers the shared `sarama.Client` owns, which is safe only because every sarama use site calls `Open` first — is correct today and commented, but it is the sharpest dependency on unspecified behaviour in the change.

**MSK broker-version compatibility is unproven.** The suites run against Confluent's image. The Fetch-version clamp against the broker's advertised ceiling should degrade correctly on an older MSK Provisioned cluster, but that path is untested until someone runs it against one. Flagged for manual verification before release.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -131,6 +131,22 @@ blocks:
             - cd cmd/ui/frontend && yarn install && yarn build && cd -
             - make test-schema-registry
 
+  # Only the main txn-discovery suite runs here. The HA suite
+  # (make test-txn-discovery-ha) is deliberately excluded: it is opt-in, slow,
+  # and inherently timing-sensitive, since it kills a broker mid-stream and
+  # asserts on recovery.
+  - name: 'integration: txn discovery'
+    dependencies: ['resolve frontend dependencies', 'resolve go dependencies']
+    task:
+      jobs:
+        - name: 'txn discovery'
+          commands:
+            - cache restore frontend-dependencies
+            - cache restore go-mod-cache
+            - cache restore go-build-cache
+            - cd cmd/ui/frontend && yarn install && yarn build && cd -
+            - make test-txn-discovery
+
   - name: 'integration: migration e2e'
     dependencies: ['resolve frontend dependencies', 'resolve go dependencies']
     task:
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ pre-commit-install: ## Install git pre-commit hooks
 # Tests
 # ==============================================================================
 
-.PHONY: test-go test-tf-validation test-playwright test-go-coverage test-go-coverage-ui test-migration test-migration-setup test-migration-teardown test-osk-scan test-kafka-connect test-schema-registry
+.PHONY: test-go test-tf-validation test-playwright test-go-coverage test-go-coverage-ui test-migration test-migration-setup test-migration-teardown test-osk-scan test-kafka-connect test-schema-registry test-txn-discovery
 
 test-go: build-frontend ## Run Go unit tests (excludes Terraform validation; see test-tf-validation)
 	go test $(GOTEST_FLAGS) ./...
@@ -125,6 +125,16 @@ test-schema-registry: build ## Run Schema Registry scan tests (unauthenticated, 
 	@bash integration-tests/schema-registry/setup.sh
 	@bash integration-tests/schema-registry/run.sh || (bash integration-tests/schema-registry/teardown.sh; exit 1)
 	@bash integration-tests/schema-registry/teardown.sh
+
+# Depends on `build`, not `build-frontend`: the suite execs ./kcp as a
+# subprocess (KTD5), so the whole binary has to exist, not just the assets the
+# unit tests embed. Teardown is a trap rather than the duplicated
+# `|| (teardown; exit 1)` of the targets above so it also runs when the test
+# binary panics or the run is interrupted, neither of which reaches a `||`.
+test-txn-discovery: build ## Run transaction-discovery E2E tests against a docker-compose broker
+	@bash integration-tests/txn-discovery/setup.sh
+	@trap 'bash integration-tests/txn-discovery/teardown.sh' EXIT; \
+	go test -tags e2e_txndiscovery -timeout 20m -v ./integration-tests/txn-discovery/...
 
 # ==============================================================================
 # State-file backward-compat archive (real generated kcp-state.json fixtures)
--- a/Makefile
+++ b/Makefile
@@ -142,8 +142,9 @@ test-txn-discovery: build ## Run transaction-discovery E2E tests against a docke
 # is the one CI runs.
 #
 # Its build tag is separate from that suite's so that `test-txn-discovery` cannot
-# accidentally pick this file up — the two need different clusters on different
-# ports and would fight over them.
+# accidentally pick this file up, and its brokers listen on 29192-29194 rather
+# than 29092, so the two stacks can coexist instead of failing at `compose up`
+# over an already-bound host port.
 test-txn-discovery-ha: build ## Run transaction-discovery leader-failover E2E tests (opt-in, 3 brokers, slow)
 	@bash integration-tests/txn-discovery-ha/setup.sh
 	@trap 'bash integration-tests/txn-discovery-ha/teardown.sh' EXIT; \
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ pre-commit-install: ## Install git pre-commit hooks
 # Tests
 # ==============================================================================
 
-.PHONY: test-go test-tf-validation test-playwright test-go-coverage test-go-coverage-ui test-migration test-migration-setup test-migration-teardown test-osk-scan test-kafka-connect test-schema-registry test-txn-discovery
+.PHONY: test-go test-tf-validation test-playwright test-go-coverage test-go-coverage-ui test-migration test-migration-setup test-migration-teardown test-osk-scan test-kafka-connect test-schema-registry test-txn-discovery test-txn-discovery-ha
 
 test-go: build-frontend ## Run Go unit tests (excludes Terraform validation; see test-tf-validation)
 	go test $(GOTEST_FLAGS) ./...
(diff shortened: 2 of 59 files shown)


adrian-januzi:
Unsure how I feel about this PR being so large but there is also no nice way to break it down for easier reviews.

adrian-januzi:
Closed until further discussion.

PR #400, opened 2026-07-29
Feat/connect clusters

pinaruyar:
Summary

  Self-managed Kafka Connect connectors were stored as a flat list on the cluster with a single shared metrics blob. This PR restructures them into connect_clusters — each keyed by its connect_rest_url, carrying its own cluster-level metrics and its own list of connectors, each with per-connector metrics. The UI gains a Connect-cluster dropdown + connector dropdown, with connector metrics nested inside the cluster metrics block. This is a types.State shape change, so it lands as schema_version 2 with full backward-compatible migration.

  Motivation

  - A single Kafka cluster can front multiple Connect clusters (e.g. a plaintext and an mTLS Connect deployment); the old flat shape couldn't represent them separately or show per-cluster metrics.
  - Metrics were aggregated worker-wide with no per-connector breakdown, so the UI couldn't attribute throughput to individual connectors.

  What changed

  State shape (schema_version 1 → 2)
  - New nested types: ConnectCluster{ connect_rest_url, metrics, connectors[] } and Connector{ name, state, config, connect_host, metrics } (replaces the flat SelfManagedConnectors).
  - KafkaAdminClientInformation.ConnectClusters []ConnectCluster.
  - Per the repo's schema-freeze rules: bumped migrate.CurrentSchemaVersion, added a 1→2 upcaster in steps.go, added migration fixtures, appended a new immutable schemaShapes entry, and 
- KafkaAdminClientInformation.ConnectClusters []ConnectCluster.
- Per the repo's schema-freeze rules: bumped migrate.CurrentSchemaVersion, added a 1→2 upcaster in steps.go, added migration fixtures, appended a new immutable schemaShapes entry, and regenerated the golden.
- Scan behavior for a given Connect cluster is unchanged (re-scan replaces that cluster's connectors); clusters coexist rather than overwriting each other.

Migration robustness (fix(state))
- Fixes a gap where era-C state files without a schema_version and carrying self_managed_connectors bypassed the upcaster and failed strict decoding. Now era-B and unversioned era-C files migrate correctly. Covered by new fixtures (era-b-…, era-c-v0-…, era-c-v1-…).

Per-connector metric collection (feat(metrics))
- Jolokia and Prometheus collectors now group source-task/sink-task metrics by connector and populate both ConnectCluster.Metrics (cluster-level) and each Connector.Metrics.
- Transient "<metric> (<connector>)" label encoding is split back out during scan so nested storage stays clean.

Prometheus client-level parity (fix(prometheus))
- Switches the four client-level queries from kafka_connect_network_io_* to kafka_connect_metrics_* (the names the JMX exporter actually produces), reaching parity with Jolokia. Requires the companion migration-tooling change that adds the client-id-aware exporter rule.

Noise reduction (fix(jmx))
- Absent-MBean warnings are de-duped (warn-once) with an ErrJolokiaMBeanNotFound sentinel, and Prometheus no-data drops to Debug — so a source-only connector no longer spams sink-metric warnings.

UI (feat(ui) + fix(ui))
- Connect-cluster dropdown + connector dropdown; connector metrics render nested inside the Connect-cluster metrics block.
- Each metrics block is keyed by cluster/connector so it remounts and re-initializes its date range on switch (fixes stale dates), and metric labels are sorted alphabetically for stable table/dropdown/chart/CSV ordering.

Backward compatibility

Old state files (v0.4.0+) migrate to the new shape transparently on read — internal/state/migrate never mutates the file. Inspect with kcp state version, migrate on disk with kcp state upgrade.

Testing

- New/updated unit tests across internal/types, internal/state/migrate, internal/services/{jmx,prometheus,report}, internal/client, and the scanner (incl. connect_metrics_split).
- Playwright e2e: new connect-clusters.spec.ts + updated fixtures.
- Schema snapshot/freeze tests updated per the append-only rules.
- Validated end-to-end on the osk-playground against two coexisting Connect clusters (plaintext via Jolokia, mTLS via Prometheus).


adrian-januzi:
Intersting observation around upgrading state files with pre-existing scanned self-managed connectors. I don't think there is much we can do about it now as we can't retroactively associate a Connect cluster's URL to a state file's entry as we never saved it anywhere.

Existing state file (`v0.6.5` I believe) that has a connector in it:
```shell
❯ jq '.regions[0].clusters[0].kafka_admin_client_information.self_managed_connectors' kcp-state.json
{
  "connectors": [
    {
      "name": "datagen-pageviews",
      "config": {
        "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
        "iterations": "10000000",
        "kafka.topic": "pageviews",
        "key.converter": "org.apache.kafka.connect.storage.StringConverter",
        "max.interval": "100",
        "name": "datagen-pageviews",
        "quickstart": "pageviews",
        "tasks.max": "1",
        "value.converter": "org.apache.kafka.connect.json.JsonConverter",
        "value.converter.schemas.enable": "false"
      },
      "state": "RUNNING",
      "connect_host": "localhost:8083"
    }
  ]
}
```

When we upgrade the state file, it _just works_ as we expect:
```shell
❯ ./kcp state upgrade --state-file kcp-state.json

Executing kcp with build version=0.0.0-localdev commit=87fa16ccf8b30e016f47f021b66ebe5bf059d581 date=2026-07-30T10:02:00Z
upgraded kcp-state.json (schema_version stamped)
```

The pre-existing connectors get moved into a `"connect_rest_url": ""` as that field is populated based on the `--connect-url` flag while previously it was just populated by the internal, and useless, connect worker URL:
```shell
❯ jq '.msk_sources.regions[0].clusters[0].kafka_admin_client_information.connect_clusters[0]' kcp-state.json
{
  "connect_rest_url": "",
  "connectors": [
    {
      "name": "datagen-pageviews",
      "state": "RUNNING",
      "config": {
        "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
        "iterations": "10000000",
        "kafka.topic": "pageviews",
        "key.converter": "org.apache.kafka.connect.storage.StringConverter",
        "max.interval": "100",
        "name": "datagen-pageviews",
        "quickstart": "pageviews",
        "tasks.max": "1",
        "value.converter": "org.apache.kafka.connect.json.JsonConverter",
        "value.converter.schemas.enable": "false"
      },
      "connect_host": "localhost:8083"
    }
  ]
}
```

Then future scans for self-managed connectors do place the connectors appropriately by their unique URL in the statefile:
```shell
❯ ./kcp scan self-managed-connectors --state-file kcp-state.json --connect-rest-url http://kcp-xxxxx.us-east-1.elb.amazonaws.com --cluster-id arn:aws:kafka:us-east-1:xxxxx:cluster/kcp-playground-public/a54f2f41xxxxx-2 --use-sasl-scram --sasl-scram-username xxxxxx --sasl-scram-password xxxxxx

Executing kcp with build version=0.0.0-localdev commit=87fa16ccf8b30e016f47f021b66ebe5bf059d581 date=2026-07-30T10:02:00Z
🚀 Starting self-managed connector scan for cluster kcp-playground-public
  🔍 Found 5 connectors
  ✅ Successfully retrieved connector details for 5 connectors
✅ Updated cluster kcp-playground-public with self-managed connector information
✅ Self-managed connector scan complete for cluster kcp-playground-public



❯ jq '.msk_sources.regions[0].clusters[0].kafka_admin_client_information.connect_clusters' kcp-state.json
[
  {
    "connect_rest_url": "",
    "connectors": [
      {
        "name": "datagen-pageviews",
        "state": "RUNNING",
        "config": {
          "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
          "iterations": "10000000",
          "kafka.topic": "pageviews",
          "key.converter": "org.apache.kafka.connect.storage.StringConverter",
          "max.interval": "100",
          "name": "datagen-pageviews",
          "quickstart": "pageviews",
          "tasks.max": "1",
          "value.converter": "org.apache.kafka.connect.json.JsonConverter",
          "value.converter.schemas.enable": "false"
        },
        "connect_host": "localhost:8083"
      }
    ]
  },
  {
    "connect_rest_url": "http://kcp-xxxxx.us-east-1.elb.amazonaws.com",
    "connectors": [
      {
        "name": "datagen-stock-trades",
        "state": "RUNNING",
        "config": {
          "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
          "kafka.topic": "smoke-stock-trades",
          "key.converter": "org.apache.kafka.connect.storage.StringConverter",
          "max.interval": "1000",
          "name": "datagen-stock-trades",
          "quickstart": "stock_trades",
          "tasks.max": "1",
          "value.converter": "org.apache.kafka.connect.json.JsonConverter",
          "value.converter.schemas.enable": "false"
        },
        "connect_host": "kafka-connect.kcp-playground-public.local:8083"
      },
      {
        "name": "datagen-orders",
        "state": "RUNNING",
        "config": {
          "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
          "kafka.topic": "smoke-orders",
          "key.converter": "org.apache.kafka.connect.storage.StringConverter",
          "max.interval": "1000",
          "name": "datagen-orders",
          "quickstart": "orders",
          "tasks.max": "1",
          "value.converter": "org.apache.kafka.connect.json.JsonConverter",
          "value.converter.schemas.enable": "false"
        },
        "connect_host": "kafka-connect.kcp-playground-public.local:8083"
      },
      {
        "name": "datagen-users",
        "state": "RUNNING",
        "config": {
          "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
          "kafka.topic": "smoke-users",
          "key.converter": "org.apache.kafka.connect.storage.StringConverter",
          "max.interval": "1000",
          "name": "datagen-users",
          "quickstart": "users",
          "tasks.max": "1",
          "value.converter": "org.apache.kafka.connect.json.JsonConverter",
          "value.converter.schemas.enable": "false"
        },
        "connect_host": "kafka-connect.kcp-playground-public.local:8083"
      },
      {
        "name": "datagen-pageviews",
        "state": "RUNNING",
        "config": {
          "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
          "kafka.topic": "smoke-pageviews",
          "key.converter": "org.apache.kafka.connect.storage.StringConverter",
          "max.interval": "1000",
          "name": "datagen-pageviews",
          "quickstart": "pageviews",
          "tasks.max": "1",
          "value.converter": "org.apache.kafka.connect.json.JsonConverter",
          "value.converter.schemas.enable": "false"
        },
        "connect_host": "kafka-connect.kcp-playground-public.local:8083"
      },
      {
        "name": "datagen-ratings",
        "state": "RUNNING",
        "config": {
          "connector.class": "io.confluent.kafka.connect.datagen.DatagenConnector",
          "kafka.topic": "smoke-ratings",
          "key.converter": "org.apache.kafka.connect.storage.StringConverter",
          "max.interval": "1000",
          "name": "datagen-ratings",
          "quickstart": "ratings",
          "tasks.max": "1",
          "value.converter": "org.apache.kafka.connect.json.JsonConverter",
          "value.converter.schemas.enable": "false"
        },
        "connect_host": "kafka-connect.kcp-playground-public.local:8083"
      }
    ]
  }
]
```

adrian-januzi:
Another thing I noticed is that when you re-run `kcp scan self-managed-connectors` but without the `--metrics` flag on the second run, you lose all the per-connector metrics previously gathered.

Run 1, **with** metrics:
```shell
❯ ./kcp scan self-managed-connectors --state-file kcp-state.json --cluster-id arn:aws:kafka:us-east-1:xxxxx:cluster/kcp-playground-public/a54f2f41xxxxx-2 --connect-rest-url http://kcp-xxxxx.us-east-1.elb.amazonaws.com --use-sasl-scram --sasl-scram-username xxxxxx --sasl-scram-password xxxxxx --metrics prometheus --metrics-range 1d --credentials-file apache-kafka-credentials.yaml

Executing kcp with build version=0.0.0-localdev commit=87fa16ccf8b30e016f47f021b66ebe5bf059d581 date=2026-07-30T10:43:28Z
🚀 Starting self-managed connector scan for cluster kcp-playground-public
  🔍 Found 5 connectors
  ✅ Successfully retrieved connector details for 5 connectors
✅ Updated cluster kcp-playground-public with self-managed connector information
  📊 Collected 17292 Connect metric data points
✅ Self-managed connector scan complete for cluster kcp-playground-public
```

Every connector gets its own metrics block, alongside the cluster-level one:
```shell
❯ jq '.msk_sources.regions[0].clusters[0].kafka_admin_client_information.connect_clusters[0]
      | {url: .connect_rest_url,
         cluster_metrics: (.metrics.aggregates | keys),
         connectors: [.connectors[] | {name, has_metrics: (.metrics != null),
                                       metrics: (if .metrics then (.metrics.aggregates | keys) else null end)}]}' kcp-state.json
{
  "url": "http://kcp-xxxxx.us-east-1.elb.amazonaws.com",
  "cluster_metrics": [
    "connector-count",
    "task-count"
  ],
  "connectors": [
    {
      "name": "datagen-stock-trades",
      "has_metrics": true,
      "metrics": [
        "source-record-poll-rate",
        "source-record-write-rate"
      ]
    },
    {
      "name": "datagen-orders",
      "has_metrics": true,
      "metrics": [
        "source-record-poll-rate",
        "source-record-write-rate"
      ]
    },
    {
      "name": "datagen-users",
      "has_metrics": true,
      "metrics": [
        "source-record-poll-rate",
        "source-record-write-rate"
      ]
    },
    {
      "name": "datagen-pageviews",
      "has_metrics": true,
      "metrics": [
        "source-record-poll-rate",
        "source-record-write-rate"
      ]
    },
    {
      "name": "datagen-ratings",
      "has_metrics": true,
      "metrics": [
        "source-record-poll-rate",
        "source-record-write-rate"
      ]
    }
  ]
}
```

Example connector metrics:
```shell
❯ jq '.msk_sources.regions[0].clusters[0].kafka_admin_client_information.connect_clusters[0].connectors[]
      | select(.name=="datagen-orders") | .metrics
      | {aggregates, data_points: (.results | length), query_info: (.query_info | length)}' kcp-state.json
{
  "aggregates": {
    "source-record-poll-rate": {
      "avg": 2.0093337375518647,
      "max": 2.3147288548900193,
      "min": 1.7175572519083968,
      "p95": null,
      "p99": null
    },
    "source-record-write-rate": {
      "avg": 2.0095642094124715,
      "max": 2.328398596667873,
      "min": 1.7177757984476396,
      "p95": null,
      "p99": null
    }
  },
  "data_points": 2882,
  "query_info": 2
}
```

Now run 2 w/ the same command without `--metrics`, as if to refresh configs:
```shell
❯ ./kcp scan self-managed-connectors --state-file kcp-state.json --cluster-id arn:aws:kafka:us-east-1:xxxxx:cluster/kcp-playground-public/a54f2f41xxxxx-2 --connect-rest-url http://kcp-xxxxx.us-east-1.elb.amazonaws.com --use-sasl-scram --sasl-scram-username xxxxxx --sasl-scram-password xxxxxx

Executing kcp with build version=0.0.0-localdev commit=87fa16ccf8b30e016f47f021b66ebe5bf059d581 date=2026-07-30T10:43:28Z
🚀 Starting self-managed connector scan for cluster kcp-playground-public
  🔍 Found 5 connectors
  ✅ Successfully retrieved connector details for 5 connectors
✅ Updated cluster kcp-playground-public with self-managed connector information
✅ Self-managed connector scan complete for cluster kcp-playground-public
```

Here we see we lose the metrics per-connector. I think we should try to preserve these metrics unless you have any objections? The main one I can think of is: `datagen-stock-trades` exists with metrics, Omar deletes it, redeploys it with lower throughput/tasks/etc. Then on a future config-only scan, we would associate the old metrics with a technically newer connector.
```shell
❯ jq -c '.msk_sources.regions[0].clusters[0].kafka_admin_client_information.connect_clusters[0].connectors[]
         | {name, has_metrics: (.metrics != null),
            metrics: (if .metrics then (.metrics.aggregates|keys) else null end)}' kcp-state.json
{"name":"datagen-stock-trades","has_metrics":false,"metrics":null}
{"name":"datagen-orders","has_metrics":false,"metrics":null}
{"name":"datagen-users","has_metrics":false,"metrics":null}
{"name":"datagen-pageviews","has_metrics":false,"metrics":null}
{"name":"datagen-ratings","has_metrics":false,"metrics":null}
```

docs/assets/apache-kafka-configuration/connect-metrics-collection.md
```diff
@@ -71,17 +71,19 @@ JVM — no additional configuration is required beyond the Jolokia agent.
...
 (`kafka_connect_worker_*`, `kafka_connect_source_task_*`) are typically exported
-by default. The client-level metrics (`kafka_connect_network_io_*`) require the
-JMX exporter to whitelist the `kafka.connect:client-id=*,type=connect-metrics`
-MBean pattern — if these are not exported, `kcp` logs a warning and continues
+by default. The client-level metrics (`kafka_connect_metrics_*`) require the
+JMX exporter to whitelist the `kafka.connect:type=connect-metrics,client-id=*`
+MBean pattern **and** a rule that captures the `client-id` key (a rule matching
+only `type=connect-metrics` with no `client-id` will not match the real
+ObjectName) — if these are not exported, `kcp` logs a warning and continues
 collecting the remaining metrics.
```
adrian-januzi, line 87: correct me if I missed it but should we add this somewhere in the command's long description so that its visible on the [command's docs page](https://confluentinc.github.io/kcp/0.8.11/command-reference/scan/self-managed-connectors/) + help usage

adrian-januzi:
wrt. https://github.com/confluentinc/kcp/pull/400#issuecomment-5129970071

Can we leave a comment in the upcaster or somewhere else in the code to just explain the scenario of when this happens and why. Just a simple two-liner or so about where the `connect_url` came from, the fact it is a useless internal URL and that we didn't save the old `--connect-url` so no way to retrieve it for proper keying in this new setup.

internal/types/self_managed_connect.go
```diff
@@ -1,13 +1,24 @@
...
 }
 
-type SelfManagedConnectors struct {
-	Connectors []SelfManagedConnector `json:"connectors"`
-	Metrics    *ConnectClusterMetrics `json:"metrics,omitempty"`
+// ConnectCluster groups the connectors discovered from one Connect REST endpoint
+// (the scan's --connect-rest-url), the stable identity of "which Connect cluster".
+// Metrics holds the cluster-level (worker/client) metrics for this endpoint.
+type ConnectCluster struct {
+	ConnectRestURL string                 `json:"connect_rest_url"`
```
adrian-januzi, line 21: I'm not sure if it's possible because I think we require some form of connector auth flag but there is a potential that an Omar passes the `--conect-url https://admin:pw@connect:8083` which might persist secrets in plaintext to the state file.

cmd/create_asset/migrate_connectors/self_managed/cmd_migrate_self_managed_connectors.go
adrian-januzi, line 1: There's no way to specify specific connect clusters tied to a Kafka cluster for this migration command, right? I think that's fine to be honest, especially when we move to a direct-api migration model, just curious.

adrian-januzi (commented):
some small nits/questions/thoughts

internal/types/self_managed_connect.go
pinaruyar: yeah auth flag is required but user still can pass both. Should we reject this specific url type?

cmd/create_asset/migrate_connectors/self_managed/cmd_migrate_self_managed_connectors.go
pinaruyar: I think it is possible to add --connect-rest-url in this command as well. Just maybe not in this PR, open to conversation

internal/types/self_managed_connect.go
adrian-januzi: I think that's possibly the easiest way to avoid this scenario: quick regex to check for `[a-z]:[a-z]@` pattern (pseudo-regex, definitely won't work lol). Just need to ensure that the user is informed in the error/rejection message as to why we rejected the URL.

I think also double-checking that that the above is the only possible auth-in-url pattern.

cmd/create_asset/migrate_connectors/self_managed/cmd_migrate_self_managed_connectors.go
adrian-januzi: Yea let's park it in backlog purgatory.

internal/types/self_managed_connect.go
pinaruyar: Design settled on rejecting just the userinfo form (`user:pass@`) — traced through why, wanted to leave the reasoning here since it directly follows from your "is this the only auth-in-url pattern" question.

Also considered rejecting secret-shaped query/fragment params (`?token=`, `?api_key=`, `#access_token=...`), but dropped it after tracing the actual request path:

- `ListConnectors()` builds the request as `fmt.Sprintf("%s/connectors", baseURL)`. If the URL already has a query string, `/connectors` lands inside the query string instead of the path — the request actually hits `/` instead of `/connectors`. A real Connect worker's root endpoint returns a JSON object, but `ListConnectors()` decodes into `[]string`, so that fails immediately, `Run()` returns early, and `SetConnectCluster`/`PersistStateFile` never execute. The URL never reaches the state file — the scan just fails with a decode error.
- Separately, kcp has no bearer-token/API-key auth for self-managed Connect at all (`ConnectAuthMethod` is just `BasicAuth`/`TLS`/`Unauthenticated`), so there's no legitimate workflow this would even be protecting.

That's the opposite of the userinfo case: `https://admin:pw@connect:8083/connectors` is valid, well-routed syntax, so a *successful* scan is exactly what persists that leak. The query-param form self-aborts before ever reaching disk, so a regex for it would carry real false-positive risk (e.g. `?auth_mode=oauth`) for a leak path that doesn't actually happen in this codebase today. Left it out, but noting it here rather than silently dropping it since it was part of your original question.

(Flagging as a follow-up: if kcp ever adds token/API-key auth for self-managed Connect, this is worth revisiting — the naive URL-concatenation in `ListConnectors()` is incidental protection, not a deliberate one, and could go away if that code ever changes.)

adrian-januzi:
I'm getting errors when running the integration tests locally:

```shell
❯ make test-kafka-connect
...
cd integration-tests/connect-scan && go test -tags integration -v ./... ; \
	  status=$? ; cd ../.. ; bash integration-tests/connect-scan/teardown.sh ; exit $status
# github.com/confluentinc/kcp/integration-tests/connect-scan [github.com/confluentinc/kcp/integration-tests/connect-scan.test]
./connect_scan_test.go:54:50: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
./connect_scan_test.go:55:41: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
./connect_scan_test.go:168:50: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
./connect_scan_test.go:169:37: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
FAIL	github.com/confluentinc/kcp/integration-tests/connect-scan [build failed]
FAIL
Stopping Connect env...
```

I don't see them failing in the CI? Is that because they aren't running in CI yet or is it only an issue on my machine and they pass fine in CI?

adrian-januzi:
## `connect_rest_url: ""` is overloaded — the legacy Connect cluster is unaddressable

---

Claude found issue. I think the easiest solution is to maybe change the legacy `connect_url`'s value to `"connect_url": "unknown_connect_url"`?

---

Follow-up on the review finding at `report_service.go:418` + `apiClient.ts:215`, with a trace and repro steps. This is reachable through the documented upgrade workflow, not just theoretical.

### Mechanism

1. `internal/state/migrate/steps.go:98` — the v3 upcaster writes `connect_rest_url: ""` as a **real persisted value** for legacy connectors. The comment there is explicit that the pre-v3 schema never recorded the REST URL and that this is an unrecoverable one-way gap.
2. `cmd/ui/frontend/src/services/apiClient.ts:215` — `if (connectRestURL) queryParams.connectRestURL = connectRestURL`. `""` is falsy, so selecting that entry drops the param off the wire entirely.
3. `cmd/ui/api/api.go:336` — `c.QueryParam("connectRestURL")` returns `""` for both "absent" and "explicitly empty". The two are indistinguishable at the handler.
4. `internal/services/report/report_service.go:418` — `if connectRestURL == "" || ...ConnectRestURL == connectRestURL`. The `""` branch means "caller didn't specify → take index 0", so it short-circuits before the real comparison can ever match the legacy entry.

### The index-0 fallback is not stable

`mergeConnectClusters` (`internal/types/discovery_common.go:160-177`) emits new entries first, then appends old entries not seen this run. So one fresh scan against a real endpoint after an upgrade moves the empty-URL entry off index 0:

```
kcp state upgrade --state-file kcp-state.json                              # connect_clusters: [{""}]
kcp scan self-managed-connectors --connect-rest-url http://connect-b:8083  # [{connect-b}, {""}]
```

"Resolves to index 0" is therefore only correct while the legacy entry happens to be first.

### Repro

Start from `internal/state/migrate/testdata/era-c-v1-self-managed-connectors.json` and run `kcp state upgrade` on it — that produces the `connect_rest_url: ""` entry. Then add a second Connect cluster in the order the merge above produces, giving the two entries distinguishable metrics:

| index | `connect_rest_url` | metrics |
| --- | --- | --- |
| 0 | `http://connect-b.internal:8083` | prometheus, `task-count` = 99 |
| 1 | `""` (legacy, upcast) | jolokia, `task-count` = 1 |

```
kcp ui --state-file kcp-state.json --port 18099
curl 'http://localhost:18099/metrics/connect/osk?sessionId=default&clusterId=c1&kind=self-managed'
```

| Request | Response |
| --- | --- |
| Legacy entry selected in the UI — param dropped per step 2 | `prometheus / 99` — **index 0, the other Connect cluster** |
| `&connectRestURL=` — passing exactly the value stored for that entry | `prometheus / 99` — **still index 0** |
| `&connectRestURL=http://nope:9999` — control | 404, so URL matching itself works |

The legacy entry is not merely mislabelled, it is **unaddressable**: the only value that could match it is `""`, and line 418 spends `""` on "unspecified". No error surfaces — the user reads connect-b's throughput under the legacy cluster's label.

One caveat on method: the second Connect cluster was added by applying `mergeConnectClusters`' ordering directly rather than standing up a live Connect endpoint for a real scan. Steps 1–4, the upcaster output, and the HTTP responses above are all from the real binary and real server.

### Suggested fix

Make "unspecified" a distinct signal from `""` — a `*string` param, or an explicit sentinel — rather than overloading the empty string to mean both "caller didn't specify" and a legitimate stored value.

pinaruyar:
> I'm getting errors when running the integration tests locally:
> 
> ```shell
> ❯ make test-kafka-connect
> ...
> cd integration-tests/connect-scan && go test -tags integration -v ./... ; \
> 	  status=$? ; cd ../.. ; bash integration-tests/connect-scan/teardown.sh ; exit $status
> # github.com/confluentinc/kcp/integration-tests/connect-scan [github.com/confluentinc/kcp/integration-tests/connect-scan.test]
> ./connect_scan_test.go:54:50: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
> ./connect_scan_test.go:55:41: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
> ./connect_scan_test.go:168:50: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
> ./connect_scan_test.go:169:37: c.KafkaAdminClientInformation.SelfManagedConnectors undefined (type "github.com/confluentinc/kcp/internal/types".KafkaAdminClientInformation has no field or method SelfManagedConnectors)
> FAIL	github.com/confluentinc/kcp/integration-tests/connect-scan [build failed]
> FAIL
> Stopping Connect env...
> ```
> 
> I don't see them failing in the CI? Is that because they aren't running in CI yet or is it only an issue on my machine and they pass fine in CI?

This does not show up in CI because apparently the integration tests have been deleted previously in a hotfix when we removed connector scan and my branches added them back but was not aware they were not running in semaphore

adrian-januzi (approved):
lgtm!

adrian-januzi (approved):

PR #405, opened 2026-08-10
feat: make `kcp migration` command hot-reload aware

adrian-januzi:
The gist:

A pod no longer requires a full cycle for all CRD updates, instead depending on the change, it can hot-reload the changes.

Change | Behavior
-- | --
configId change only | Hot-reload
Modify an existing route in place (same position, same port) | Hot-reload
Add or remove a route | Roll
Rename a route that has TLS or authentication configured | Roll
Streaming domain metadata (name/type/bootstrap endpoint) | Hot-reload
Streaming domain nodeIdRanges | Roll
New TLS secret introduced (route or streaming domain) | Roll (reusing an already-mounted secret does not roll)
Admin/static config changes | Roll

For a hot-reload change, we can no longer rely on waiting for the pod to cycle to ensure the change is applied (because no pods cycle). Therefore, we rely on a `spec.configId` change which we inject into the CR we are about to apply. Then we poll the `/config` endpoint on the Gateway to validate that the pods now reflect the `configId` we passed before apply.

---

To hot-reload or not:

To determine if hot-reloads are enabled **and possible** (as we cannot rely on just `gateway.hotReload.enabled` being present as not every change we apply is a hot-reload), we have some 'pre-flight' checks:
1. At `init` (both `kcp migration init` and FSM state `initialised`) we check if the field `gateway.hotReload.enabled` is set. This tells us if we even have an alternative to pod rolls.
2. At `init`, we check if the `/config` endpoint and field `spec.configId` are present. This confirms that if hot-reloads are possible, we can set the change identifier and verify the changes.
3. The `gateway.hotReload.enabled` and `spec.configId` state/mutability stays consistent between all three CRDs.
4. Before the `fence` state and block clients, we pass a `spec.configId` to the CR to ensure that the field is definitely writable.

This results in 3 behaviours:
- `RolloutOnly`: matches previous behaviour of always cycling the pods on every CR-change transition.
- `PerPodConfigId`: polls the `/config` endpoint always making no assumptions if a change might be a hot-reload or pod roll.
- `error`: Reverts to `initialised` state of FSM, when possible (not mid-migration), to allow traffic to continue to the source.

This ensures that regardless of setup, kcp will always validate a change has been applied successfully or failed.


cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -97,6 +99,8 @@ the migration state file and must be provided each time.`,
 	optionalFlags.BoolVar(&insecureSkipTLSVerify, "insecure-skip-tls-verify", false, "Skip TLS certificate verification for REST endpoint and Kafka connections.")
 	optionalFlags.StringVar(&clusterRestCaCert, "cluster-rest-ca-cert", "", "Path to a CA certificate that verifies the destination cluster REST endpoint's TLS certificate. Use when the REST endpoint is HTTPS behind a private/internal CA; omit for Confluent Cloud (public CA).")
 	optionalFlags.DurationVar(&rolloutTimeout, "rollout-timeout", 0, "Maximum time to wait for the Confluent operator to report the gateway as Ready during fence and switchover. 0 (the default) means no deadline — the wait runs until the operator converges or the user cancels.")
+	optionalFlags.DurationVar(&hotReloadTimeout, "hot-reload-timeout", 0, "Maximum time to wait for every gateway pod to report the new config revision when the gateway supports hot-reload. Unlike --rollout-timeout this is never unbounded: a hot-reload moves no Kubernetes signal, so 0 (the default) uses the built-in 90s budget rather than waiting forever.")
```
adrian-januzi, line 102: This definitely needs to be lowered from the default of 90s. From my own personal testing, albeit limited, I saw kcp reach times of as high as ~14 seconds for a hot-reload to successfully apply by polling the `/config` endpoint.

Will confirm with gateway team if they have any recommendations as to the above time out limit or more aggressive polling.

integration-tests/migration-hot-reload/manifests/templates/gateway-base.yaml
```diff
@@ -0,0 +1,94 @@
...
+spec:
+  replicas: __REPLICAS__
+  image:
+    # 1.3.0 is the first release serving GET /config. 1.2.x has the
+    # GATEWAY_LICENSES hook but no endpoint to verify against.
+    application: confluentinc/confluent-gateway-for-cloud:__GATEWAY_TAG__
+    init: confluentinc/confluent-init-container:__INIT_TAG__
+  # kcp overwrites spec.configId on every apply. The value here is only the
+  # starting revision, and is deliberately NOT one kcp would generate (kcp-<hex>)
+  # so a test asserting the id changed cannot pass by coincidence.
```
adrian-januzi, line 41: At the moment, the integration tests running the full e2e suite for the hot-reload work rely on either being built-from-master or from a private, development ECR. Therefore, we can't currently provide a tag into the YAMLs and we can't add the e2e integration tests to Semaphore.

integration-tests/migration-hot-reload/README.md
```diff
@@ -0,0 +1,156 @@
+# Gateway hot-reload E2E
+
+Proves, against a real licensed Confluent Gateway, that kcp confirms a gateway
+transition by **config revision** rather than by pod turnover.
```
adrian-januzi, line 4: To run these tests even locally, a Gateway license **IS** required.

internal/services/gateway/crd_contract_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+// These tests pin kcp's two hard-coded assumptions about the Gateway CRD against
+// the schema the operator actually enforces, taken from the shipped chart rather
+// than from the contract prose:
+//
+//  1. spec.configId lives at .spec.versions[].schema.openAPIV3Schema.properties
+//     .spec.properties.configId — the path crdSupportsConfigID walks.
+//  2. Generated ids satisfy the CRD's own maxLength and pattern. If they did not,
+//     every apply would be rejected by the API server, and only against the real
+//     schema is that visible without a cluster.
```
adrian-januzi, line 22: Max length 64, pattern `^[a-zA-Z0-9._:-]+$` (`alphanumeric` plus `.`, `_`, `:`, `-`). Base64 padding characters (`+`, `/`, `=`) are rejected - re-encode base64-derived IDs (e.g. hex, or base64url without padding).

internal/services/gateway/rollout_mechanism.go
```diff
@@ -0,0 +1,134 @@
...
+// status.observedGeneration caught up, so the operator has finished reconciling
+// the generation we applied. This window covers only the residual gap between
+// the operator writing that status and writing the Deployment — not "did
+// anything happen at all", which the generation baseline answers outright.
+//
+// The duration is not measured. It is the same 10s the previous detection window
+// used, kept rather than replaced with a second unmeasured number.
+//
+// var so tests can shorten it.
+var gatewayRollConfirmationWindow = 10 * time.Second
```
adrian-januzi, line 53: I believe we might have some wiggle room to lower this further and likely should.

internal/services/gateway/capability_test.go
```diff
@@ -0,0 +1,613 @@
...
+		assert.True(t, got.CRDSupportsConfigID)
+		assert.True(t, got.HotReloadEnabled)
+		assert.Empty(t, got.Advisory, "no advisory when hot-reload verification is available")
+	})
+
+	// The CFK version window that makes this a refusal rather than a downgrade:
+	// chart 0.1718.10 declares spec.hotReload.enabled but not spec.configId. Hot
+	// reload therefore works, no pod ever rolls, and rollout verification would
+	// pass having observed nothing.
+	t.Run("hot-reload on but CRD lacks configId - refuses rather than downgrading", func(t *testing.T) {
```
adrian-januzi, line 282: This is part of one of the `pre-flight` checks mentioned - `gateway.hotReload.enabled` may be settable but `spec.configId` is not/cannot. This can happen in the current release of Gateway in which hot-reloads are EA but lacks the `spec.configId` schema field which kcp relies on for confirming if a hot-reload was successful or not. The behaviour there is that we block the run as we have no way of monitoring the success/failure of a new CR update if the change was hot-reloadable.

adrian-januzi:
Parked for the moment on the basis that hot-reloads are not GA yet and are missing the `spec.configId` (and possibly the `/config` endpoint) that we need for confirming a successful hot-reload.

adrian-januzi:
### Claude's Hot-reload convergence measurement (timeout/window values)

(it was adamant these should be posted even if they are a bit useless due to being on a local rig just making small changes on the Gateway w/ no traffic through it)

Measured actual hot-reload convergence on the local minikube redundant-auth rig to justify the two defaults touched here (`73914c2e`). 40 reloads — a fresh `spec.configId` (± a `fence.scope` toggle), waiting for the gateway pod to log `Hot-reload successful {configId=…}`.

| Change type | n | min | median | mean | p90 | max |
|---|---|---|---|---|---|---|
| configId-only | 20 | 3.91 | 4.34 | 4.48 | 5.21 | **6.03** |
| fence toggle (real spec change) | 20 | 3.67 | 4.20 | 4.28 | 4.83 | **4.90** |

(seconds, patch → applied). **40/40 converged, zero pod churn** (Deployment generation unmoved). The JVM in-place apply is **~0.01s** — the entire cost is operator canary + ConfigMap promotion + kubelet projection. configId-only does trigger a real hot-reload on this build.

**Decision: keep both values.**

- **`DefaultHotReloadTimeout` = 90s** — this is a *ceiling, not a wait*: `WaitForGatewayConfigID` returns the instant every pod reports the new configId, so 90s is only ever spent on failure. Convergence is ~4–6s local / ~14s EKS, so 90s is generous headroom; lowering it buys nothing on the happy path and only makes a false failure (mid-fence, traffic blocked) more likely.
- **`gatewayRollConfirmationWindow` = 10s** — legacy-only. The modern configId path detects a roll inline from the generation baseline and never observes this window; it only applies to gateways that can't report a configId, where shrinking it risks reading a late roll as "no roll" with no configId backstop.

Env: minikube (8C/16G, k8s 1.35), CFK v0.1838.0, gateway `1.4.0-master-430`, single replica, Enterprise licence (`GTWY-CCADDON`). Note this build serves no `/config` HTTP endpoint, so the pod log line is used as the convergence signal (same event the `/config` probe observes). Harness + raw results kept locally under `~/gateway-testing/`.

ifeeneyconfluent (approved):

PR #406, opened 2026-08-11
Add Trivy vulnerability scan to pre-commit hook

pinaruyar:
Mirrors the Semaphore CI trivy fs scan so devs catch HIGH/CRITICAL vulns locally before pushing. Skips with a warning if trivy isn't installed, same as the existing golangci-lint check.
trivy isn't installed, same as the existing golangci-lint check.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -5,3 +5,9 @@ if command -v golangci-lint >/dev/null 2>&1; then
 else
   echo "warning: golangci-lint not installed, skipping Go lint"
 fi
+
+if command -v trivy >/dev/null 2>&1; then
+  trivy fs --scanners vuln --show-suppressed --severity HIGH,CRITICAL --exit-code 1 . || exit 1
+else
+  echo "warning: trivy not installed, skipping vulnerability scan"
+fi
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -173,10 +173,16 @@ minutes.
 # Install golangci-lint
 brew install golangci-lint
 
+# Install trivy
+brew install trivy
+
 # Run Go linters
 make lint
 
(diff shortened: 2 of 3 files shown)


adrian-januzi:
Thoughts on adding a section inside the `CONTRIBUTING.md` for this as we already have a section about the pre-hooks?

https://github.com/confluentinc/kcp/blob/main/CONTRIBUTING.md#linting--pre-commit-hooks

There's also a `make pre-commit-install` so we might either want to add it as part of them or a new `make pre-commit-trivy-install`? idk

pinaruyar:
> There's also a `make pre-commit-install` so we might either want to add it as part of them or a new `make pre-commit-trivy-install`? idk
I have already added trivy as part of the `make pre-commit-install` I think a separate install seems redundant? Do you possibly mean adding 'make trivy' on top of the `make pre-commit-install`?
@adrian-januzi

adrian-januzi:
@pinaruyar I'm not seeing the change to the Makefile - is it pushed up?

Let's leave it as part of the `make pre-commit-install`.

pinaruyar:
> I'm not seeing the change to the Makefile - is it pushed up?
> Let's leave it as part of the `make pre-commit-install`.

https://github.com/confluentinc/kcp/blob/cf09e3370dae2640d401248eb91962f7c0f96864/Makefile#L73-L74
I might be wrong but the make file refers to the .githooks/pre-commit which already has the changes in so that should be enough for `make pre-commit-install` to pick up the new trivy @adrian-januzi

adrian-januzi:
Oh I'm sorry that's my bad, I didn't realise that's how the Make command works. LGTM then

adrian-januzi (approved):

adrian-januzi (approved):

PR #408, opened 2026-08-11
feat(migration): add hidden --run-report flag to migration execute

bmartin494:
## What this adds

A hidden `--run-report <path>` flag on `kcp migration execute`. When set, it writes
per-stage migration timings to a JSON file. When not set, nothing changes at all.

`execute` already walks a state machine — `wait_for_lags → fence → pause_offset_sync →
verify_fence → promote → switch`. This records how long each transition took:

```json
{
  "migration_id": "migration-7847bcf6-…",
  "kcp_version": "0.0.0-localdev",
  "kcp_commit": "c257cbe6…",
  "topics": 10,
  "lag_threshold": 400,
  "duration_ms": 374845,
  "stages": [
    { "event": "wait_for_lags", "from": "initialized", "to": "lags_ok",
      "started": "…", "ended": "…", "duration_ms": 877 },
    { "event": "fence", "from": "lags_ok", "to": "fenced",
      "started": "…", "ended": "…", "duration_ms": 174109 }
  ],
  "final_state": "switched",
  "outcome": "completed"
}
```

A failed stage is recorded with `failed: true` and the error text, and the file is
flushed after **every** stage rather than at the end — so a run killed mid-flight still
yields everything that completed. That matters because `CheckLags` and `PromoteTopics`
have no convergence bound of their own, so an external supervisor killing the process is
a normal outcome, not an exceptional one.

## Why

We are building an end-to-end live migration test against real infrastructure — two
Confluent Cloud Dedicated clusters, a real cluster link, EKS, CFK, a real Confluent
Gateway, and real client load through it — intended as a **release gate**: run it against
a build to answer "did this break the migration?"

The gate needs per-stage timings, not just a total. Two examples of what that bought
during development:

- A run stalled at the fence for its full 15-minute rollout timeout. The report showed
  `fence` at 900,963ms with `failed: true` and the rollout error, which is what
  identified the cause.
- Across three runs on separate environments, `promote` came in at 37.099s / 36.935s /
  37.135s and `switch` at 162.638s / 162.417s / 162.724s. That tightness makes these
  useful regression detectors — a build that moved `promote` to 60s would stand out
  immediately. It is also only visible at stage granularity.

## The honest alternative, and why not

**These timings are already derivable from `kcp.log`** by differencing the stage banners.
Checked against a real run:

| Stage | From log banners | From run report |
| --- | --- | --- |
| `wait_for_lags` | 2s | 2.149s |
| `fence` | 174s | 174.404s |
| `promote` | 43s | 42.767s |
| `switch` | 163s | 162.719s |

So this flag is a convenience, not a capability. The case for it is:

1. **No coupling to human-readable log strings.** A reworded banner or changed emoji
   silently breaks a scraper — and breaks it into wrong timings rather than a clean
   error. For an automated gate, where a false failure is worse than no gate, taking a
   dependency on log prose is the wrong trade.
2. **Structured outcome data** — `final_state`, `outcome`, per-stage `failed`/`error`,
   `skipped_stages` — rather than inferring them.
3. **Provenance.** `kcp_version` and `kcp_commit` recorded by kcp itself, which for a gate
   testing a specific build is the whole point.
4. Millisecond resolution. The 0.2s reproducibility above is invisible at 1s granularity.

If reviewers would rather the test rig scraped the log, that is a legitimate position and
this can be dropped — the rig would just carry the coupling instead.

## Footprint

```
run_report.go                212 ++    new, self-contained
run_report_test.go           268 ++    tests — nearly half the diff
atomic_write.go               46 ++    extracted from state.go
orchestrator.go               20 +-    one field, one setter, 4 call sites
migration_executor.go         23 +-    construct + defer Finish
cmd_migration_execute.go      10 ++    the flag
state.go                      38 --    SHRANK — now uses the shared helper
```

The four orchestrator call sites are on a **nil-tolerant** recorder: with no
`--run-report`, every method is a no-op and behaviour is byte-identical. `state.go` got
smaller because the atomic-write helper it already had was extracted and shared.

Stage timing is taken in `Execute`'s loop rather than in FSM callbacks, deliberately —
looplab's non-reentrant `eventMu` makes doing work inside a callback hazardous, and the
existing code already avoids it for rollback events.

## Why hidden

Following the existing precedent for `kcp migrate`, which is registered but hidden. This
is tooling support rather than a user-facing feature, so it is excluded from `--help` and
from generated docs, while still being covered by tests so a refactor cannot quietly
break it.

Happy to unhide and document it instead if reviewers prefer — the argument for that is
that a release gate depending on an undocumented flag is a slightly odd contract.

## Testing

- `run_report_test.go` — 268 lines covering stage recording, failure capture, skipped
  stages, incremental flush, nil-recorder tolerance and atomic write.
- `make test-go` green: 71 packages ok, 0 failed.
- Exercised against real infrastructure: four migrations reached `switched` on live
  Confluent Cloud clusters, plus the failure paths described above (a stalled fence
  rollout, and a run killed by an external deadline mid-stage).

### DIFF
--- a/cmd/migration/execute/cmd_migration_execute.go
+++ b/cmd/migration/execute/cmd_migration_execute.go
@@ -41,6 +41,7 @@ var (
 	detectUnroutedProducersDuration time.Duration
 	consumerOffsetSyncDrainDuration time.Duration
 	promoteBatchSize                int
+	runReport                       string
 )
 
 func NewMigrationExecuteCmd() *cobra.Command {
@@ -100,6 +101,7 @@ the migration state file and must be provided each time.`,
 	optionalFlags.IntVar(&promoteBatchSize, "promote-batch-size", 0, "Maximum number of mirror topics to promote per batch. 0 (the default) promotes all topics at once. When set (>0), each batch is promoted and confirmed STOPPED before the next batch is submitted.")
 	optionalFlags.DurationVar(&detectUnroutedProducersDuration, "detect-unrouted-producers-duration", 0, "Time to monitor source offsets after fencing to detect producers still writing directly to the source cluster (bypassing the gateway); a detected increase aborts the migration before switchover. 0 (the default) skips the check; minimum 10s if set.")
 	optionalFlags.DurationVar(&consumerOffsetSyncDrainDuration, "consumer-offset-sync-drain-duration", 0, "How long to wait after fencing before disabling the cluster link's consumer.offset.sync.enable. The fence freezes source consumer offsets, so this drain lets the link propagate the final offsets to the destination, reducing (best-effort, not guaranteed) messages reprocessed after switchover. Has no effect unless the migration was initialised with --pause-consumer-offset-sync. 0 (the default) disables the wait.")
+	optionalFlags.StringVar(&runReport, "run-report", "", "Write per-stage migration timings to <path> as JSON.")
 	migrationExecuteCmd.Flags().AddFlagSet(optionalFlags)
 	groups[optionalFlags] = "Optional Flags"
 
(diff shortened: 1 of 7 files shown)


adrian-januzi (approved):
lgtm

PR #410, opened 2026-08-13
feat(migration): drive kcp migration from a declarative config

adrian-januzi:
Replaces the `kcp migration` flag surface with a single declarative `gateway-migration.yaml`. **64 flags → 10.**

```
flags   ──┐
          ├──► MigrationConfig ──► migration-state.json (unchanged) ──► existing path
-f YAML ──┘
```

## Scope boundaries

This is **pass 1 = parity**, and it holds three hard lines:

| Boundary | Meaning |
|---|---|
| `migration-state.json` is not modified | Same shape, same field names, same absence of a `schema_version`. Its three existing pin-tests stay green untouched — that is the scope boundary's own test. |
| No new capability the flags did not already give | The manifest is a *second deserializer into the same struct*, not a feature. No topic globs, no `include`/`exclude`, no CP-on-six-auth-methods, no `clusterId` self-discovery. |
| The manifest is the page's canonical shape | Not a superset, not a tidier variant. |

One deliberate departure from strict parity, forced rather than chosen:

- **`--skip-validate` × `--pause-consumer-offset-sync` mutual exclusion is dropped** — a cobra flag cannot be grouped with a YAML field, so the constraint becomes inexpressible. It is not safety-bearing: the snapshot is already taken unconditionally, `offset_sync_bookend.go` has an explicit empty-snapshot fallback, and `workflow.go` already guards the snapshot independently of the CLI rule (with tests that say in as many words that they pin it *in case the CLI rule goes away*).

## Resulting surface

| Command | Flags after |
|---|---|
| `kcp migration init` | `-f/--file`, `--migration-state-file`, `--skip-validate` |
| `kcp migration execute` | `-f/--file`, `--migration-state-file`, `--migration-id`, `--accept-spec-change` |
| `kcp migration lag-check` | `-f/--file`, `--poll-interval` |
| `kcp migration list` | `--migration-state-file` (unchanged) |

**The verbs are unchanged.** An earlier revision of this branch renamed `execute` → `apply` with `execute` kept as a hidden deprecated alias; `dfcaac92` reverts that in full, so `ffc92ce1`'s commit message describes a rename that is no longer in the tree. There is no alias and no deprecation notice — `execute` is the single registered verb, as before.

Of the 10, ~8 are retained flags; `-f` and `--accept-spec-change` are new. The retained ones are all **addressing and escape hatches** — which file, which state file, which migration, how often to poll, skip validation. Nothing describing the migration itself remains a flag.

## What the flag surface was hiding

1. **Source auth was declared twice and never cross-checked.** `init` registered six `--use-*` bools and the credential strings, then never contacted the source — those strings were written and *read nowhere*. `execute` re-declared the same six and actually consumed them. **This PR fixes the declared-twice half and not the changed-between-runs half**: credentials are still never persisted, so there is nothing to compare a changed auth block against. Please don't let the release notes imply otherwise.
2. **18 flags were declared twice and had drifted** — `init` lacked `--sasl-scram-mechanism` and `--aws-region`, so an IAM-authenticated MSK source passed `init` and then discovered at `execute` that a region was required.
3. **Naming inconsistency** — `lag-check` called it `--rest-endpoint`, `init` called the same thing `--cluster-rest-endpoint`. Dissolves into one YAML field.

Each has a cheaper direct fix, so the config file is **not** justified by defect removal. The case is ergonomics — 64 → 10 — plus not obstructing the PRFAQ's desired-state model.

## Design decisions worth reviewing

- **`kind: GatewayMigration`** alongside the existing `kind: Migration`, sharing `apiVersion`, sub-types and parser. `ParseKind` dispatch means pointing `migration init -f` at a `kcp migrate` manifest yields "this is a `kcp migrate` manifest", not a field-by-field unknown-key dump.
- **`CredentialsRef` — credentials inline or referenced.** A polymorphic string-or-mapping, so a mounted Kubernetes Secret still works as a path. Note goccy does not propagate `yaml.Strict()` into a custom unmarshaler; the inner decode re-applies it.
- **`${ENV_VAR}` interpolation, opt-in per file** via top-level `interpolate: true`. Inlining credentials makes the manifest secret-bearing, so interpolation is a *requirement* of this change, not a nicety. Resolution runs **after** parse and **before** validation: substituting into raw bytes is a YAML-injection hazard (a password containing a newline or `: ` rewrites the document), and running after validation yields errors like `ca_cert file "${CA_PATH}": no such file`.
- **Drift check.** The state file's topology snapshot stays authoritative; a manifest that disagrees is a hard error, overridable with `--accept-spec-change` only once the cutover is past the point where re-running `init` is safe. A warning would not do — everyone arriving from `kubectl` or Terraform expects a desired-state file to reconcile, and a line that scrolls past during an irreversible cutover is not consent. Terminal output is section names and counts only, never a topic list. **This is the weakest part of pass 1** and is honestly the thing most worth a reviewer's scepticism: users get a file that looks like desired state and are told that after `init`, half of it is read-only.
- **Durations are `time.Duration` strings** (`10m`), deliberately a second spelling next to `kind: Migration`'s `IntervalMs int`. The retired flags were `DurationVar`s, so this is the parity-preserving choice; the conversion of `IntervalMs` is parked.
- **Second generated schema** (`gatewaymigration.schema.json`) from one `//go:generate` that writes both, so they cannot drift apart, with a staleness guard test.
- **Every field ported from a retired flag carries that flag's usage text as its schema `description`** — retiring ~56 flags also retires ~56 pieces of `--help` guidance, and for a CLI run from a bastion `--help` is the primary discovery surface.

## Verification

Full Minikube e2e run twice, the second time after the final security fixes so the evidence matches the committed tree:

```
23/23 top-level, 94 including subtests, 0 failures, ~22 min, exit 0
ok  github.com/confluentinc/kcp/integration-tests/migration  1339.703s
```

Plus `make test-go` 74 packages ok; `golangci-lint run ./...` 0 issues; `gofmt` and `go vet` clean in both tag modes.

## Out of scope, deliberately

- **Phase 6 doc debt is outstanding.** `docs/assets/getting-started-with-zero-cut-migrations.md` still documents `--auth-mode`, `--promote-batch-size` and other retired flags, and `CHANGELOG.md` is untouched. The phasing table says phase 6 ships with 3–5, so **this branch currently ships stale docs** — worth deciding before merge.
- No e2e coverage for `spec.topics` exact-match, for the drift hard-error, or for `lag-check`; and the fixture's single principal makes the destination Kafka vs. REST credential split unobservable end to end.


cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -1,311 +1,338 @@
...
-			if usage != "" {
-				fmt.Printf("%s:\n%s\n", groupNames[i], usage)
-			}
-		}
-
-		fmt.Println("All flags can be provided via environment variables (uppercase, with underscores).")
+	cmd.Flags().StringVarP(&manifestFile, "file", "f", "", "Path to the GatewayMigration manifest describing this migration.")
+	cmd.Flags().StringVar(&migrationStateFile, "migration-state-file", "", "Path to the migration-state.json file (produced by kcp migration init).")
+	cmd.Flags().StringVar(&migrationId, "migration-id", "", "Address a migration by id instead of by the manifest's metadata.name. Needed only for migrations registered before metadata.name became the identity.")
+	cmd.Flags().BoolVar(&acceptSpecChange, "accept-spec-change", false, "Proceed even though the manifest no longer matches the topology snapshot taken at init. Only meaningful once the cutover is past the point where re-running init is safe.")
```
adrian-januzi: Still on the fence with `--accept-spec-change` where a drift between the `migration.yaml` and `migration-state.json` exists, the `migration.yaml` is preferred as the source of truth.

Do we want to give Omar this escape hatch?

cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -1,311 +1,338 @@
...
+	// here rather than dereferencing on an invariant that lives elsewhere,
+	// because the alternative failure is a nil panic mid-cutover.
+	if dstCreds.SASLPlain == nil {
+		return MigrationExecutorOpts{}, fmt.Errorf("spec.target.kafka.credentials: only sasl_plain is supported for the destination in this release")
+	}
+
+	// Policy is read FRESH on every run and never snapshotted. Only these two
+	// reach the config; the rest were already purely execute-time.
+	config.DetectUnroutedProducersDuration = g.Spec.Policy.DetectUnroutedProducersDuration
+	config.ConsumerOffsetSyncDrainDuration = g.Spec.Policy.ConsumerOffsetSyncDrainDuration
```
adrian-januzi: Unsure about this - these flags were always set on the `kcp migration execute` command so in a sense they were always 'read FRESH on every run' which we will still honour now.

However, maybe they should be snapshotted to the `migration-state.yaml` along with the rest of the infra topology though that means the user needs to apply them at `kcp migration init` which might be odd and potentially locks them into that setup for all topic-migration batches for that migration set?

cmd/migration/init/cmd_migration_init.go
```diff
@@ -70,268 +51,202 @@ after client traffic has already been fenced.
...
-
-		for i, fs := range flagOrder {
-			usage := fs.FlagUsages()
-			if usage != "" {
-				fmt.Printf("%s:\n%s\n", groupNames[i], usage)
-			}
-		}
-
-		fmt.Println("All flags can be provided via environment variables (uppercase, with underscores).")
+	migrationInitCmd.Flags().StringVarP(&manifestFile, "file", "f", "", "Path to the GatewayMigration manifest describing this migration.")
```
adrian-januzi: Thoughts, opinions, feelings on having the flag just a simple `-f` / `--file`? I prefer it because it is easier to write quickly but typically in the rest of the codebase, our flags have been way more verbose.

internal/manifest/gateway.go
```diff
@@ -0,0 +1,539 @@
...
+	if mc.SASLPlain != nil {
+		return nil
+	}
+	if mc.IAM == nil && mc.SASLScram == nil && mc.MTLS == nil &&
+		mc.UnauthenticatedTLS == nil && mc.UnauthenticatedPlaintext == nil {
+		return nil // no block at all — the shared validator reports that
+	}
+	return []error{fmt.Errorf(
+		"spec.target.kafka.credentials: only sasl_plain is supported for the destination in this release")}
+}
```
adrian-januzi, line 381: This is currently true on main too - we only accept SASL/PLAIN over TLS for the destination cluster which covers Confluent Cloud and specific CP setups.

This should be a follow-up item.

adrian-januzi: Main defaulting to CC setup: https://github.com/confluentinc/kcp/blob/main/cmd/migration/execute/migration_executor.go#L214-L216

cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -1,311 +1,364 @@
...
+	// external migration performance rig (its only consumer) drives it this way.
+	runReport string
 )
 
+// reversibleStates are the states with nothing irreversible behind them. Drift
+// found here is answered with "re-run init", which is cheap and re-validates
+// the new spec properly. Past these, producers are fenced and/or offset sync is
+// disabled, and re-running init would discard what completes or rolls back the
+// cutover — so the only way forward is an explicit --accept-spec-change.
+var reversibleStates = []string{
```
ifeeneyconfluent: Hmm concerned about this knowledge of which states are reversable lives in the cobra command layer.  Feels like this is something the state machine itself should be in control of.

adrian-januzi: [I'm still on the fence of even keeping this flag if I'm totally honest.](https://github.com/confluentinc/kcp/pull/410#discussion_r3782771098)

I like the idea of providing a escape-hatch but at the same time, what's it really escaping? A user not having to re-run `kcp migration init` again?

ifeeneyconfluent (commented):
Automated review findings (10 items, most severe first). 9 are left as inline comments below; one more couldn't be anchored inline since it's outside this PR's diff hunks:

**security — `cmd/migrate/apply/cmd_migrate_apply.go` (`runApply`, near the manifest load at the top of the function)**: `kind: Migration`'s credential fields became secret-bearing `CredentialsRef` in this PR, but `runApply` loads the manifest via bare `os.ReadFile`+`manifest.Parse` and never calls the world/group-readable permission warning that `gateway.go` has for the identical hazard.

Failure scenario: an operator inlines a `sasl_scram` block directly into a kcp-migrate manifest saved with default 0644 permissions. `kcp migrate apply`/`validate` give no warning, while the sibling GatewayMigration path warns loudly for the identical new hazard.

internal/manifest/migration.go
```diff
@@ -71,31 +71,44 @@ type Spec struct {
...
+	ClusterID        string         `yaml:"clusterId,omitempty" json:"clusterId,omitempty"`
+	Kafka            *TargetKafka   `yaml:"kafka,omitempty" json:"kafka,omitempty"`
+	SchemaRegistry   *Endpoint      `yaml:"schemaRegistry,omitempty" json:"schemaRegistry,omitempty"`
+	Connect          *Endpoint      `yaml:"connect,omitempty" json:"connect,omitempty"`
 }
 
+// TargetKafka is the destination Kafka cluster. RestEndpoint is the only field
+// kind: Migration uses; the rest are optional additions for kind:
+// GatewayMigration, which also dials the destination bootstrap directly.
 type TargetKafka struct {
```
ifeeneyconfluent, line 100: **correctness**: `TargetKafka.Credentials`/`RestCredentials` are accepted by `kind: Migration` manifests but silently ignored at apply time.

An author copies `spec.target.kafka.credentials: ./creds.yaml` from the new GatewayMigration docs into a `kind: Migration` manifest; it parses/validates cleanly but apply (`cmd_migrate_apply.go:382-384`) authenticates via `Target.ClusterCredentials` instead, with no warning the field did nothing.

internal/manifest/gateway.go
```diff
@@ -0,0 +1,541 @@
...
+func ParseGatewayMigration(data []byte) (*GatewayMigration, error) {
+	kind, err := ParseKind(data)
+	if err != nil {
+		return nil, err
+	}
+	if kind != KindGatewayMigration {
+		return nil, wrongKindError(kind, KindGatewayMigration)
+	}
+
+	var g GatewayMigration
```
ifeeneyconfluent, line 183: **correctness**: `ParseGatewayMigration` runs `parseStrict` (which decodes `time.Duration` natively) before `interpolateInto`, so env-var interpolation on duration fields is architecturally impossible.

A manifest with `interpolate: true` and `spec.policy.rolloutTimeout: "${ROLLOUT_TIMEOUT}"` fails at parse time with a duration-format error instead of substituting the env var; no test covers interpolation on a duration field.

internal/manifest/schemagen/schemagen.go
```diff
@@ -37,6 +51,157 @@ func Generate() ([]byte, error) {
 		}
 	}
 
+	if err := addCredentialDefs(s); err != nil {
+		return nil, err
+	}
+	// Every credentials slot is a CredentialsRef: a path string or an inline
+	// mapping of the referenced file's shape.
+	polymorphic(source.Properties["credentials"], defMigrateCredentials)
```
ifeeneyconfluent, line 59: **correctness**: `Generate()` never calls `polymorphic()` on `target.kafka.credentials`/`restCredentials` the way `GenerateGateway()` does for its schema.

Verified in generated `migration.schema.json`: `spec.target.kafka.credentials` is emitted as a raw object requiring both `Path` and `Inline` with `additionalProperties:false`, rejecting every valid spelling (plain path string) the real parser accepts.

internal/manifest/schemagen/schemagen.go
```diff
@@ -37,6 +51,157 @@ func Generate() ([]byte, error) {
...
+	gateway := spec.Properties["gateway"]
+	policy := spec.Properties["policy"]
+	clusterLink := spec.Properties["clusterLink"]
+
+	source.Properties["type"].Enum = []any{manifest.SourceMSK, manifest.SourceApacheKafka}
+	target.Properties["type"].Enum = []any{manifest.TargetConfluentCloud, manifest.TargetConfluentPlatform}
+
+	if err := addCredentialDefs(s); err != nil {
+		return nil, err
+	}
```
ifeeneyconfluent, line 106: **correctness**: `GenerateGateway()` doesn't patch `kafka.Required`, so the schema only requires `restEndpoint` even though `Validate()` also requires non-empty `bootstrapServers` and `credentials`.

A manifest omitting `spec.target.kafka.bootstrapServers` or `.credentials` validates cleanly against the published schema (false confidence in editor/CI lint) and is only caught when `kcp migration init` actually runs.

internal/manifest/gateway.go
```diff
@@ -0,0 +1,541 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if errs := checkRestIsAPIKeyForm(*tc); len(errs) > 0 {
+			return nil, errs[0]
+		}
+		return tc, nil
+	}
+
+	mc, errs := g.DestinationKafkaCredentials()
```
ifeeneyconfluent, line 467: **efficiency**: `RestCredentials()` calls `DestinationKafkaCredentials()` internally when `restCredentials` is omitted, and both `init` and `execute` commands call both methods separately in the same run, double-reading the same file.

On a Kubernetes projected secret volume mid-rotation, the two reads could observe different content, causing the SASL/PLAIN broker credential and derived REST credential to silently diverge — a hard-to-root-cause auth mismatch.

cmd/migration/init/cmd_migration_init.go
```diff
@@ -70,268 +51,204 @@ after client traffic has already been fenced.
...
+	// and the pre-disable link-config snapshot and strand a live cutover.
+	//
+	// This runs BEFORE credentials are resolved: it is a safety refusal, and an
+	// operator re-running init mid-cutover needs to hear "this migration is
+	// fenced" rather than have it masked by an unset environment variable.
+	if err := checkReInitIsSafe(migrationState, g.Metadata.Name); err != nil {
+		return err
+	}
+
+	if err := checkCredentialsResolve(g); err != nil {
```
ifeeneyconfluent: **correctness**: `checkCredentialsResolve` runs unconditionally before the `--skip-validate` early-return and `os.Stat`s local mTLS/TLS cert/key paths.

Running `kcp migration init --skip-validate` from a bastion/CI host without the source cluster's mTLS cert mounted fails with 'file not found', even though the codebase elsewhere anticipates deferring such validation to a host that has the cert (KubeConfigPath split); no test covers this case.

cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -1,311 +1,364 @@
...
+// It compares only the topology fields. Credentials are out of scope
+// automatically — they are never persisted. Policy is out of scope by omission,
+// not by absence: two policy fields (the detect-unrouted and offset-sync-drain
+// durations) DO reach the state file via MigrationConfig, but detectDrift never
+// reads them back, because policy is re-read fresh from the manifest every run
+// so the snapshot's copy is never authoritative. Two fields are additionally
+// excluded on purpose: the kubeconfig path, because execute is resume-safe and
+// may legitimately run from a different machine or pod; and anything
+// credential-bearing, because credentials are never persisted.
+func detectDrift(g *manifest.GatewayMigration, config *migration.MigrationConfig) []string {
```
ifeeneyconfluent, line 234: **simplification**: `detectDrift` is a manual per-field comparison between manifest and state-file snapshot rather than a generalized structural diff.

A new topology-relevant field added to `GatewaySpec` later silently escapes drift detection unless someone remembers to add a matching comparison line.

internal/manifest/schemagen/schemagen.go
```diff
@@ -37,6 +51,157 @@ func Generate() ([]byte, error) {
...
+	polymorphic(source.Properties["credentials"], defMigrateCredentials)
+	kafka := target.Properties["kafka"]
+	polymorphic(kafka.Properties["credentials"], defMigrateCredentials)
+	polymorphic(kafka.Properties["restCredentials"], defTargetCredentials)
+
+	// Durations parse as "10m", not as an integer count.
```
ifeeneyconfluent, line 124: **test-coverage**: The hardcoded 3-name list of duration fields patched into the schema is mirrored exactly by the guard test's list, so the test can't catch an omission.

A future `time.Duration` field added to `Policy` that isn't added to both lists gets reflected as `{"type":"integer"}` while the real parser expects a duration string, and no test flags the mismatch.

internal/manifest/gateway.go
```diff
@@ -0,0 +1,541 @@
...
+	// Fenced is a local FILE path, snapshotted into the state file at init.
+	Fenced string `yaml:"fenced" json:"fenced"`
+	// Switchover is a local FILE path, snapshotted into the state file at init.
+	Switchover string `yaml:"switchover" json:"switchover"`
+}
+
+// Policy is the execute-time knobs. Every value is optional and zero carries
+// meaning: 0 promotes all at once / imposes no deadline / skips the check.
+type Policy struct {
+	LagThreshold     int `yaml:"lagThreshold,omitempty" json:"lagThreshold,omitempty"`
```
ifeeneyconfluent, line 113: **correctness**: `Policy.LagThreshold` is `omitempty` and only validated as non-negative, never required, whereas the retired `--lag-threshold` flag was `MarkFlagRequired`.

A YAML indentation or key-name typo under `policy:` now silently falls back to `lagThreshold=0` instead of erroring the way the missing required flag used to.

internal/manifest/migration.go
adrian-januzi: Need to look into this further but I think this is about @tjunderhill's direct API migration work as those YAML's are named `kind: Migration` where as I've differentiated the ones as part of this work by labelling them `kind: GatewayMigration` though will confirm shortly.

adrian-januzi: Example above was specifically about the `kcp migrate` (Tom's, this is gonna get confusing) command but the premise of the problem spanned both as they shared `TargetKafka`.

internal/manifest/gateway.go
adrian-januzi: Slightly hesitant to touch this as it seems like a super niche edge-case --> Not doing.

cmd/migration/init/cmd_migration_init.go
adrian-januzi: Fixed in 609c2aca: moved `checkCredentialsResolve` (both source and destination-Kafka legs) into Phase 5, alongside `RestCredentials()`, so it now runs after the `--skip-validate` early return. `--skip-validate` defers credential resolution symmetrically for both legs rather than singling out the source leg for eager validation — which matters as both grow more auth methods (mTLS certs, etc.) that may need local file access to resolve.

Added `TestInit_SkipValidateSkipsCredentialResolution` to lock in the new behavior, and updated `TestInit_RejectsInvalidSourceCredentials` / `TestInit_RejectsIAMOnApacheKafkaSource` to assert their errors via the non-skip path (the check still runs there, before any gateway/K8s contact).

internal/manifest/gateway.go
adrian-januzi: Fixed in e587f6e7: added `policy.Required = []string{"lagThreshold"}` in `schemagen.go` (same manual-override pattern already used for `spec.target.kafka`), so the JSON Schema now requires the key when `spec.policy` is present — catching a typo'd/missing key at lint time instead of silently defaulting to 0.

Left `Policy.LagThreshold`'s Go tag as `omitempty` and `Validate()` unchanged: 0 stays a legitimate, fail-safe value at runtime (team call), this only tightens editor/CI schema validation. Added an assertion to `TestGenerateGateway_RequiredSets` and regenerated `gatewaymigration.schema.json`.

cmd/migration/execute/cmd_migration_execute.go
adrian-januzi: Drop flag

cmd/migration/execute/cmd_migration_execute.go
adrian-januzi: Rename YAML field to `defaultPolicies` or something similar. Allow values to be overridden by flags on `kcp migration execute`.

cmd/migration/init/cmd_migration_init.go
adrian-januzi: Be consistent, rename to `--migration-yaml` or similar.

ifeeneyconfluent (approved):
LGTM

PR #414, opened 2026-08-19
feat(migration): derive the gateway fence from the initial CR, not a fenced-CR file

adrian-januzi:
## What & why

`kcp migration` no longer takes a user-supplied **fenced gateway CR file**. At fence time kcp reads the stored **initial CR snapshot**, injects `fence: {scope: ALL, errorCode: BROKER_NOT_AVAILABLE}` onto the route(s) named in `spec.gateway.fence.routes`, and applies the patched CR.

This means there is just a single YAML the user has to provide and less overall chance of drift between the initial CRs state and the final switchover CRs state.

### Manifest change

```yaml
spec:
  gateway:
    crs:
      initial: migration-gateway              # unchanged (k8s object name)
      switchover: /etc/kcp/gateway-switchover.yaml
    fence:
      routes:
        - migration-route                      # NEW — crs.fenced is removed
```

## How it works

- **`cleanInitialCR`** — extracted from `unfenceGateway` (metadata strip), now the single canonical base for both the fence (`FenceRoutes(cleanInitialCR(initial))`) and both fence-removal paths (unfence re-applies it; switchover applies the fence-free switchover CR).
- **`ValidateGatewayCRs`** — drops the fenced CR; instead confirms the named routes **exist in the initial CR** and are **not already fenced**, and keeps the switchover-lifts-the-fence checks.
- **State** — `MigrationConfig.FencedCrYAML []byte` → `FenceRoutes []string`. Confirmed outside the `types.State` schema freeze (separate file/struct); `internal/types` stays green.
- **Drift** — `fence.routes` drifts as an order-insensitive **set**; reported as a bare flag with counts only, never route names.

## Out of scope

The **switchover** CR stays file-based (a larger transform — repoints routes to the destination domain and drops the source route). Fence params (`scope: ALL`, `errorCode: BROKER_NOT_AVAILABLE`) are hardcoded; making them configurable later is additive.


docs/assets/migration-assets/gateway-examples/gateway-migration.yaml
```diff
@@ -93,8 +93,15 @@ spec:
...
-      fenced: /etc/kcp/gateway-fenced.yaml # a local FILE path
       switchover: /etc/kcp/gateway-switchover.yaml # a local FILE path
+    # Route(s) to fence at cutover, by spec.routes[].name. kcp reads the live
+    # initial CR, injects fence: {scope: ALL, errorCode: BROKER_NOT_AVAILABLE}
+    # onto each named route, and applies the patched CR — there is no separate
+    # fenced-CR file to create or keep in sync. The named routes must exist in
+    # the initial CR and must not already be fenced.
+    fence:
+      routes:
+        - migration-route
```
adrian-januzi, line 104: For the `migration.yaml`, this is all the user now has to do for kcp to apply a fence to their route during a switchover.

One question though: should we limit this to a single string? Would there be a case where we want to fence multiple routes in an 'all-at-once' migration (not thinking about TBR/TBM yet)?

internal/services/gateway/fence.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, fmt.Errorf("parsing base gateway CR: %w", err)
+	}
+	return FenceRoutesObj(obj, routeNames)
+}
+
+// FenceRoutesObj is FenceRoutes for a base CR the caller has already parsed
+// into a plain map[string]any (e.g. migration.cleanInitialCR), sparing a
+// redundant parse/marshal round trip when the caller already holds the tree.
+// obj is mutated in place with the injected fence block(s).
+func FenceRoutesObj(obj map[string]any, routeNames []string) ([]byte, error) {
```
adrian-januzi, line 40: In-place fencing to the initial CR happens here

internal/services/gateway/validate.go
```diff
@@ -86,13 +86,19 @@ func (s *K8sService) ValidateGatewayCRs(ctx context.Context, namespace, gatewayN
...
 // Split from the method so unit tests can inject a fake clientset, matching
 // waitForGatewayPods and waitForGatewayReady.
-func validateGatewayCRs(ctx context.Context, clientset kubernetes.Interface, namespace, gatewayName string, initialYAML, fencedYAML, switchoverYAML []byte) (CRValidationResult, error) {
+//
+// There is no fenced CR to check: the fence is derived at cutover from the live
+// initial CR by injecting a fence block onto the routes named in fenceRoutes
+// (see gateway.FenceRoutes). So instead of cross-checking a fenced file, this
+// confirms those routes exist in the initial CR and are not already fenced, then
+// that the switchover CR lifts the fence rather than leaving them blocked.
+func validateGatewayCRs(ctx context.Context, clientset kubernetes.Interface, namespace, gatewayName string, initialYAML, switchoverYAML []byte, fenceRoutes []string) (CRValidationResult, error) {
```
adrian-januzi, line 101: Slowly building up our validation on the initial CR 👀

internal/services/gateway/fence.go
```diff
@@ -0,0 +1,130 @@
...
+	return patched, nil
+}
+
+// FencedRouteNames returns the spec.routes[].name values that carry a
+// non-nil fence block — the inverse of FenceRoutes. It exists to recover
+// route names from a pre-38f5c974 fenced-CR snapshot (the file FenceRoutes
+// replaced): a migration-state.json written by that older kcp has no
+// fence_routes to resume from, and this reconstructs it from the fenced CR it
+// did snapshot. See MigrationConfig.UnmarshalJSON in the migration package.
+func FencedRouteNames(fencedCRBytes []byte) ([]string, error) {
```
adrian-januzi: Part of me thinks we should drop this. It seems like a redundant and unlikely path to even accommodate.

adrian-januzi: Just dropped it - not realistically needed.

ifeeneyconfluent (approved):

PR #415, opened 2026-08-19
feat(osk): override broker metric names for relabelled exporters

adrian-januzi:
## Summary

Apache Kafka (OSK) users whose Prometheus exporter relabels the standard broker
series — or whose Jolokia agent exposes non-standard MBean names — could not
collect the affected metrics without patching kcp source: the built-in queries
matched nothing and silently returned empty results. This adds per-cluster
metric-name overrides in `apache-kafka-credentials.yaml`, so those users repoint
kcp at the names their infrastructure actually exposes — no source changes, no
recording rules required.

## What's new

Two optional maps, keyed on the seven logical broker-metric labels
(`BytesInPerSec`, `BytesOutPerSec`, `MessagesInPerSec`, `PartitionCount`,
`GlobalPartitionCount`, `ClientConnectionCount`, `TotalLocalStorageUsage`):

| Config key                  | Backend    | Overrides                          |
| --------------------------- | ---------- | ---------------------------------- |
| `prometheus.metric_names`   | Prometheus | base PromQL series name            |
| `jolokia.mbean_overrides`   | Jolokia    | MBean object name                  |

```yaml
clusters:
  - id: my-cluster
    prometheus:
      url: http://prom:9090
      metric_names:
        BytesInPerSec: acme_broker_bytesin_total
    jolokia:
      endpoints: [ ... ]
      mbean_overrides:
        BytesInPerSec: "acme.kafka:type=BrokerTopicMetrics,name=BytesInPerSec"
```

## Design decisions

- **Overrides key on the logical label, not the default series name** — stable,
  backend-agnostic, and identical to the `label` already written to the state
  file, so the same key works for both backends.
- **Prometheus is a base-series-name rename, not a full-PromQL rewrite** — kcp
  keeps its existing wrapping (`sum(rate(…))`, `sum(…)`, the GiB conversion) and
  the overridden name is threaded through `PrometheusMetric`, so `filter.labels`
  injection keeps working. `GlobalPartitionCount` preserves its
  `{name="GlobalPartitionCount"}` discriminator on top of the renamed series.
- **Unknown keys are a loud load-time error, not a silent no-op** — a single
  canonical label set in `internal/types` backs the validator; drift tests in
  the Prometheus and JMX packages assert their definitions cover exactly that
  set, so a validated key can never silently fail to apply. Only the (non-secret)
  override keys are echoed in the error; the override values are not.
- **An overridden-but-still-empty metric is elevated from DEBUG to WARN** — the
  override exists precisely to fix an empty result, so a still-empty result is
  actionable rather than routine noise.

No `types.State` shape change, so no schema-version bump (the credentials structs
are not part of persisted state).

## Validation

- `make test-go` — full suite green (0 failures, 74 packages). New tests cover
  override construction, byte-identical no-override regression, label/validator
  drift guards, the WARN-elevation for empty overridden metrics on both backends,
  and load-time rejection of unknown keys.
- `golangci-lint run ./...` — 0 issues. `go.mod`/`go.sum` unchanged.
- CLI exercised end-to-end: a credentials file with a misspelled override key
  fails fast with
  `metric_names: unknown metric label(s) BytesInPersec, NotARealMetric; valid labels are: …`
  (exit 1); a file with valid keys passes credential validation and proceeds to
  the scan.

> **Note:** this was a `strike` run — no pipeline review preceded it, so this PR
> review is the first review.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Claude Code

### DIFF
--- a/internal/services/prometheus/prometheus_service.go
+++ b/internal/services/prometheus/prometheus_service.go
@@ -25,18 +25,49 @@ type MetricQuery struct {
 	// GroupByConnector indicates the query is aggregated with `sum by (connector) (...)`
 	// and results should be broken out per connector using the `connector` series label.
 	GroupByConnector bool
+	// Overridden is true when PrometheusMetric came from a user-configured
+	// metric-name override rather than the default series name. An overridden
+	// query that still returns no data is actionable (the override was set
+	// precisely to fix an empty result), so it is logged louder than a routine
+	// empty default.
+	Overridden bool
 }
 
 // BrokerQueryDefinitions returns the standard Kafka broker Prometheus queries.
-func BrokerQueryDefinitions() []MetricQuery {
+// overrides maps a logical label (e.g. "BytesInPerSec") to the base series name
+// this cluster's exporter actually exposes; an entry with an empty value is
+// ignored. The overridden name is substituted into kcp's existing query wrapping
+// and set as PrometheusMetric, so applyLabelFilter injection keeps working. Pass
+// nil for the defaults.
+func BrokerQueryDefinitions(overrides map[string]string) []MetricQuery {
+	name := func(label, def string) (string, bool) {
+		if v, ok := overrides[label]; ok && v != "" {
+			return v, true
+		}
+		return def, false
+	}
+
+	bytesIn, bytesInOv := name("BytesInPerSec", "kafka_server_brokertopicmetrics_bytesinpersec_total")
+	bytesOut, bytesOutOv := name("BytesOutPerSec", "kafka_server_brokertopicmetrics_bytesoutpersec_total")
+	messagesIn, messagesInOv := name("MessagesInPerSec", "kafka_server_brokertopicmetrics_messagesinpersec_total")
+	partitionCount, partitionCountOv := name("PartitionCount", "kafka_server_replicamanager_partitioncount")
+	globalPartition, globalPartitionOv := name("GlobalPartitionCount", "kafka_controller_kafkacontroller_value")
+	clientConn, clientConnOv := name("ClientConnectionCount", "kafka_server_socketservermetrics_connection_count")
+	logSize, logSizeOv := name("TotalLocalStorageUsage", "kafka_log_log_size")
+
+	// GlobalPartitionCount is distinguished by a {name="..."} discriminator on a
+	// shared controller series; an override replaces the base series name and the
+	// discriminator is preserved.
+	globalPartitionSel := globalPartition + `{name="GlobalPartitionCount"}`
+
 	return []MetricQuery{
-		{Label: "BytesInPerSec", Query: "sum(rate(kafka_server_brokertopicmetrics_bytesinpersec_total[%s]))", PrometheusMetric: "kafka_server_brokertopicmetrics_bytesinpersec_total"},
-		{Label: "BytesOutPerSec", Query: "sum(rate(kafka_server_brokertopicmetrics_bytesoutpersec_total[%s]))", PrometheusMetric: "kafka_server_brokertopicmetrics_bytesoutpersec_total"},
-		{Label: "MessagesInPerSec", Query: "sum(rate(kafka_server_brokertopicmetrics_messagesinpersec_total[%s]))", PrometheusMetric: "kafka_server_brokertopicmetrics_messagesinpersec_total"},
-		{Label: "PartitionCount", Query: "sum(kafka_server_replicamanager_partitioncount)", PrometheusMetric: "kafka_server_replicamanager_partitioncount"},
-		{Label: "GlobalPartitionCount", Query: "kafka_controller_kafkacontroller_value{name=\"GlobalPartitionCount\"}", PrometheusMetric: "kafka_controller_kafkacontroller_value{name=\"GlobalPartitionCount\"}"},
-		{Label: "ClientConnectionCount", Query: "sum(kafka_server_socketservermetrics_connection_count)", PrometheusMetric: "kafka_server_socketservermetrics_connection_count"},
-		{Label: "TotalLocalStorageUsage", Query: "sum(kafka_log_log_size) / (1024*1024*1024)", PrometheusMetric: "kafka_log_log_size"},
+		{Label: "BytesInPerSec", Query: "sum(rate(" + bytesIn + "[%s]))", PrometheusMetric: bytesIn, Overridden: bytesInOv},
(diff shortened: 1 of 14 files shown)


internal/services/prometheus/prometheus_service.go
```diff
@@ -25,18 +25,51 @@ type MetricQuery struct {
...
-		{Label: "BytesOutPerSec", Query: "sum(rate(kafka_server_brokertopicmetrics_bytesoutpersec_total[%s]))", PrometheusMetric: "kafka_server_brokertopicmetrics_bytesoutpersec_total"},
-		{Label: "MessagesInPerSec", Query: "sum(rate(kafka_server_brokertopicmetrics_messagesinpersec_total[%s]))", PrometheusMetric: "kafka_server_brokertopicmetrics_messagesinpersec_total"},
-		{Label: "PartitionCount", Query: "sum(kafka_server_replicamanager_partitioncount)", PrometheusMetric: "kafka_server_replicamanager_partitioncount"},
-		{Label: "GlobalPartitionCount", Query: "kafka_controller_kafkacontroller_value{name=\"GlobalPartitionCount\"}", PrometheusMetric: "kafka_controller_kafkacontroller_value{name=\"GlobalPartitionCount\"}"},
-		{Label: "ClientConnectionCount", Query: "sum(kafka_server_socketservermetrics_connection_count)", PrometheusMetric: "kafka_server_socketservermetrics_connection_count"},
-		{Label: "TotalLocalStorageUsage", Query: "sum(kafka_log_log_size) / (1024*1024*1024)", PrometheusMetric: "kafka_log_log_size"},
+		{Label: "BytesInPerSec", Query: "sum(rate(" + bytesIn + "[%s]))", PrometheusMetric: bytesIn, Overridden: bytesInOv},
```
pinaruyar, line 66: `bytesIn` here is the user-supplied override value (from `prometheus.metric_names`), concatenated straight into a template that still contains `%s` for the rate window — and `CollectMetrics` later does `query = fmt.Sprintf(query, rateWindow)` on the result (line 209) whenever the built query still contains the substring `%s`. Only the override *keys* are validated (`validateMetricNameOverrides`), never the values, so a stray `%` in someone's metric name gets misread as a format verb:

- `metric_names: {BytesInPerSec: "foo%dbar"}` → query becomes `sum(rate(foo%!d(string=5m)bar[%!s(MISSING)]))` — Prometheus rejects it outright.
- More dangerous — `metric_names: {PartitionCount: "weird%s_count"}` (see line 69): `PartitionCount`'s template (`sum(weird%s_count)`) has no *intended* placeholder, but `strings.Contains(query, "%s")` still matches on the override text itself, so `Sprintf` silently substitutes the rate window in: `weird%s_count` → `weird5m_count`. No error, no warning — kcp just queries a different series than the one the user configured. This is exactly the "silently wrong value" failure mode the docs warn about, except caused by this implementation detail rather than a real counter/gauge mismatch.

Suggest interpolating the rate window with something like `strings.Replace(query, "%s", rateWindow, 1)` instead of a value-dependent `Sprintf`, so an override value can never be read as a format string.

internal/services/prometheus/prometheus_service.go
```diff
@@ -121,6 +154,17 @@ func applyLabelFilter(query, metricName string, labels map[string]string) string
 	return query[:afterName] + "{" + labelStr + "}" + query[afterName:]
 }
 
+// appendNameDiscriminator appends a {name="value"} label matcher to a
+// Prometheus series name, merging it into an existing selector if the series
+// already carries one (e.g. from a metric-name override that points at a
+// job-scoped series) rather than producing a second, invalid brace group.
+func appendNameDiscriminator(series, name string) string {
+	if idx := strings.Index(series, "{"); idx >= 0 {
+		return series[:idx+1] + fmt.Sprintf(`name="%s",`, name) + series[idx+1:]
```
pinaruyar, line 163: If a `GlobalPartitionCount` override's own selector already carries a `name=` matcher, this produces two `name=` matchers in the same brace group instead of replacing/merging:

`metric_names: {GlobalPartitionCount: 'my_controller_metric{name="OtherName",job="x"}'}` →
`my_controller_metric{name="GlobalPartitionCount",name="OtherName",job="x"}`

That's syntactically valid PromQL but semantically matches zero series (a label can't equal two different values at once in one matcher). It'll fall into the existing "overridden, no data" WARN path, but the log message won't hint that it's specifically a duplicate-`name`-matcher conflict — which would save real debugging time for whoever hits it. Might be worth detecting an existing `name=` matcher here and replacing it (or rejecting the override at validation time) rather than blindly prepending a second one. Not covered by the current test suite (only the bare-series and non-`name`-selector cases are tested).

pinaruyar (commented):
Ran this branch through a full review plus a real e2e pass (`go build`/`go vet`, `make test-go`, and `make test-osk-scan` against Docker) — everything is green and the override feature works as designed (verified `TestOSKScanPrometheusMetricNameOverride` etc. actually query the relabelled series, not just "no error"). Two things I'd fix before merge, plus a lower-priority one (the first two are inline below, this one isn't since the file isn't touched by this diff):

**Jolokia MBean overrides reach an unescaped URL builder.** `internal/client/jolokia_client.go:96,151,221` build request URLs with `fmt.Sprintf("%s/read/%s", c.baseURL, mbeanPath)` — no `url.PathEscape`. That file isn't part of this PR, but this PR is what first makes `mbeanPath` user-controlled (via `jolokia.mbean_overrides`) instead of always being one of a few hardcoded MBean constants. Example: `jolokia.mbean_overrides: {BytesInPerSec: "kafka.server:type=BrokerTopicMetrics,name=BytesInPerSec?target=http://internal-host/"}` gets concatenated straight into the request URL, so a `?`/`#`/`../` in the override changes what Jolokia's own proxy actually receives. Since `mbean_overrides` values now come from a YAML file rather than fixed strings, worth adding `url.PathEscape(mbeanPath)` there (or at least a charset check in `validateMetricNameOverrides`).

pinaruyar:
out of scope but shall we extend this for the Connect metrics as well?

adrian-januzi:
> out of scope but shall we extend this for the Connect metrics as well?

@pinaruyar we could though I'd like to leave that for a follow-up PR and get this merged into main + released sooner rather than later.

pinaruyar (approved):

pinaruyar:
personally I don't think format string bug is a problem we will come across easily. Would like to think that clients use somewhat of a normal naming. I have only let Claude post them here for awareness. @adrian-januzi

PR #416, opened 2026-08-20
docs(migration): add GatewayMigration manifest reference

adrian-januzi:
## What

PR #410 introduced a declarative manifest (`gateway-migration.yaml`, `kind: GatewayMigration`) that now drives `kcp migration init|execute|lag-check`, but shipped only an example file — no docs explained how to build one.

Adds:
- **`docs/assets/gateway-manifest-reference/README.md`** — field-by-field reference: execution model, every `spec` section, credential slots (path vs. inline, `${ENV_VAR}` interpolation), the per-command flag surface, and validation rules.
- **`docs/assets/gateway-manifest-reference/gateway-migration-example.md`** — a dedicated page rendering the annotated `gateway-migration.yaml` inline via a `pymdownx.snippets` include, so the doc can't drift from the real file.


adrian-januzi:
This does not cover the changes in #414 so this should probably be merged first, then have #414 updated with docs examples and then finally merged.

adrian-januzi:
Also, tests are failing - will have a look shortly though one is still that flaky connect integration test.

ifeeneyconfluent (approved):

PR #417, opened 2026-08-20
fix(connect-scan): stop the integration test flaking randomly in CI

pinaruyar:
## Problem

The `kafka connect scan` Semaphore job fails ~25–30% of runs across **unrelated PRs and on `main` itself** — an environmental flake, not the code under test. Runtimes are bimodal: pass ≈ 9 min, fail ≈ 23 min. PR #413's curl timeout only addressed a sliver, so the failure kept relocating (it recurred e.g. on the docs-only PR #416, job `e1940f44`).

Evidence from 22 recent runs + job logs (`sem logs`/Semaphore API): failures split into two root causes.

## Root causes & fixes (keeps the real containerised test)

**1. Memory contention.** One agent runs a broker + four Connect JVMs with unbounded heaps → out of memory → job hard-killed mid-startup (died ~23 min with no test output and no epilogue flushed).
- Cap every JVM heap via `KAFKA_HEAP_OPTS` in `docker-compose.yml`.
- Move the block to the larger `s1-prod-ubuntu24-04-amd64-2` agent already used by the migration e2e block.

**2. Readiness races.**
- `wait_rest` polled `/`, which answers *before* the herder joins the group, so `POST /connectors` raced into 404s. It now gates on `GET /connectors` → 2xx (real herder-ready).
- `create_connector` captured `code=$(curl --max-time 5)` under `set -euo pipefail`, so a timed-out POST (curl exit 28) aborted the whole script instead of retrying — the observed `make … Error 28`. It now uses the `set -e`-safe `if curl -sf` idiom with an idempotent GET fallback.

**Also hardened**
- Bound + retry the in-container Jolokia download (was an unbounded `curl` that could hang a worker's startup).
- `Makefile` target now **always tears down**, **dumps `docker compose logs` on failure**, and runs `go test -timeout 8m` — so a failure is clean and diagnosable instead of a silent 23-minute stall.

## Verification

Local `make test-kafka-connect` passes with the changes — all auth subtests plus the Jolokia metrics subtest (`ok … 34.4s`), clean teardown, no leaked containers. Elimination of the flake is by construction (bounded memory can't OOM; the readiness gate reflects reality; failures are now logged); it'll be confirmed by CI over several runs, and any recurrence will now surface the container-level cause via the new logs dump.

## Note for reviewers

The agent bump to `amd64-2` is for headroom; the heap caps bound memory on their own, so if the default `amd64-1` is already large that one line can be dropped. The identical fragile Makefile pattern in `test-osk-scan` / `test-schema-registry` likely flakes the same way and could get the same treatment separately.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -122,6 +122,13 @@ blocks:
   - name: 'integration: kafka connect scan'
     dependencies: ['resolve frontend dependencies', 'resolve go dependencies']
     task:
+      agent:
+        machine:
+          # This job runs a broker + four Connect JVMs concurrently. The pipeline
+          # default (amd64-1) is memory-tight for five JVMs and was contributing to
+          # runs being hard-killed mid-startup; use the larger machine already used by
+          # the migration e2e block. (Per-JVM heaps are also capped in docker-compose.yml.)
+          type: s1-prod-ubuntu24-04-amd64-2
       prologue:
         commands:
           # keytool (from JDK) is required for TLS certificate generation
--- a/Makefile
+++ b/Makefile
@@ -126,9 +126,18 @@ test-osk-scan: build ## Run OSK scan tests (all auth methods, JMX, Prometheus)
 	  status=$$? ; cd ../.. ; bash integration-tests/osk-scan/teardown.sh ; exit $$status
 
 test-kafka-connect: build ## Run Kafka Connect self-managed connector scan tests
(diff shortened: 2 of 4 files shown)


pinaruyar:
connect scan job passed not failed - looks like it failed because I wanted to double run the pipeline therefore I manually stopped it

ifeeneyconfluent (approved):

PR #421, opened 2026-08-28
fix(migration): honour destination REST auth methods in lag-check

adrian-januzi:
## What

`kcp migration lag-check` only honoured `api_key`/`basic` on the destination REST leg. It passed `restCreds.APIKey`/`APISecret` explicitly and built its HTTP client from a bespoke `NewRESTHTTPClient` that carried TLS trust only (CA cert / insecure-skip) — so a destination that resolved to **bearer** or **mtls** (the forms #420 added) was silently downgraded: the token and the mTLS client certificate were never sent.

This routes lag-check through the same resolution the rest of `kcp migration` already uses:

- `clusterlink.Config.Auth = restCreds.Authenticator()` — carries whichever form the manifest resolved (`basic`/`bearer`/`mtls`/`api_key`), replacing the `APIKey`/`APISecret` pair.
- `restCreds.HTTPClient()` — same TLS trust as before **plus** a client certificate when the form is `mtls`, replacing `migration.NewRESTHTTPClient`.

With nothing left calling it, `migration.NewRESTHTTPClient` and its test are deleted.

lag-check still reads only the destination REST leg (`kafka.restCredentials`); it never dials the source or destination Kafka legs.

Stacks on #420 (destination auth for `init`|`execute`) → #418 → #405.


ifeeneyconfluent (dismissed):

tjunderhill (approved):
LGTM!

adrian-januzi:
weird, dunno why `auto-merge` didn't work there 🤔

PR #424, opened 2026-09-02
docs: clarify Confluent Platform support as an Apache Kafka source

adrian-januzi:
## What

Clarifies that Confluent Platform is a supported migration source (via the existing "Apache Kafka" source type), and fixes a `--source-type` inconsistency.

## Changes

- Docs: name Confluent Platform explicitly in the docs landing page, source-compatibility matrix, root README, apache-kafka-configuration guides, zero-cut migrations guide, and `scan clusters` / `create-asset migration-infra` command help.
- Fix: `scan self-managed-connectors --source-type` now accepts `apache-kafka` (not `osk`), matching every other command.

## Why

Users have been asking if CP is supported — it is, but nothing in the docs said so.


cmd/scan/self_managed_connectors/cmd_scan_self_managed_connectors.go
```diff
@@ -71,7 +71,7 @@ func NewScanSelfManagedConnectorsCmd() *cobra.Command {
     --state-file kcp-state.json \
     --connect-rest-url http://connect:8083 \
     --cluster-id my-cluster \
-    --source-type osk \
+    --source-type apache-kafka \
```
adrian-januzi, line 74: This was missed at one point, oops.

pinaruyar:
do we wanna change osk to apache kafka for the frontend as well or in files like this? https://github.com/confluentinc/kcp/blob/1eedb1b2169a77233279190543a658772e1a0a84/cmd/ui/api/api.go#L300
https://github.com/confluentinc/kcp/blob/1eedb1b2169a77233279190543a658772e1a0a84/cmd/ui/api/api.go#L303

adrian-januzi:
I did not ask Claude to post this just to check if there was more user-facing OSK references...

---

Good catch to flag, but I'd scope that out of this PR. This one only touches the user-facing `--source-type` CLI flag (already `apache-kafka` everywhere) and docs. `osk` inside `api.go`/the frontend is the *internal* wire-format token (`types.SourceTypeOSK`), not the CLI flag value, and it's more entangled than it looks:

- a static route segment: `/metrics/osk/:clusterId`
- a validated path param: `/metrics/connect/:sourceType`
- two JSON body fields: `source_type` on the ACL and mirror-topics migration-script requests
- the `sources[].type` field returned by `GET /state`

The frontend mirrors that same `'osk'` string in ~20 files (a `SourceType = 'msk' | 'osk'` union type, a `SOURCE_TYPES.OSK` constant, plus inline literals across components/hooks/store/apiClient). Renaming it means changing both sides in lockstep across all of those touchpoints, and `types.SourceTypeOSK` is shared with `scan`/`report`/`create-asset`/`healthcheck` too, so it'd ripple beyond just the UI.

Worth doing, but as its own follow-up rather than growing this docs PR — I'll open a ticket for it.

pinaruyar (approved):
looking gooddd

pinaruyar (approved):

PR #426, opened 2026-09-03
feat(migration): reshape GatewayMigration manifest to spec.topicGroup

adrian-januzi:
## What

Reshape the `GatewayMigration` manifest so one file supports both the current static (all-at-once) migration and the future topic-based one, and regenerate the IDE schemaguard to match.

- `spec.topics` + `gateway.routes` → a single `spec.topicGroup` entry.
- `gateway.crs.initial` → flattened `gateway.cr-name`.
- Bootstrap server id is no longer authored — derived from the live gateway CR at init.
- Migration mode is read from the live CR's route binding, not the manifest.

```yaml
gateway:
  namespace: confluent
  cr-name: gateway-initial
topicGroup:
  - topics: [t1, t2]          # topics and/or topicPatterns; at least one required
    topicPatterns: []         # anchored regex; ['.*'] = every mirror topic
    route: migration-route
    targetStreamingDomain: cc
```

## Why

The static and topic-based workflows need the same manifest shape. Deriving the id and mode from the live CR keeps both modes uniform (name-only reference) and removes a value operators had to look up and keep in sync.

## Breaking change

No manifest versioning: existing manifests using `spec.topics` / `gateway.routes` / `crs` now fail the strict decode with `unknown field`. Migrate-all is now an explicit `topicPatterns: ['.*']` (the omit-topics sentinel to declare 'migrate all topics on link' is gone).

## Notes

- Dynamic (topic-based) routes are refused at init — engine is out of scope here.
- Tests: `go test ./...` 74 packages pass; `golangci-lint ./...` clean; golden schema regenerated.


adrian-januzi:
One thing I want to note:

The `spec.topicGroup.topicPatterns` hasn't been fleshed out yet as I feel that should belong to the reconciliation engine and I didn't want to duplicate effort/add redundant code to a WIP piece of work.

Also, with that, the union of `spec.topicGroup.topics` and `spec.topicGroup.topicPatterns` hasn't been properly implemented either as again I didn't want to deal with the expansion of the regex in this PR as it felt out of scope when the reconciliation engine will be worked on in parallel.

internal/services/migration/workflow_test.go
```diff
@@ -75,14 +75,57 @@ func TestWorkflow_Initialize_Success(t *testing.T) {
...
 	assert.Len(t, config.ClusterLinkTopics, 3)
 	assert.Equal(t, "broker:9092", config.ClusterLinkConfigs["bootstrap.servers"])
 }
 
+// TestWorkflow_Initialize_ReusesPreFetchedCR proves Initialize does not
+// re-fetch the CR live when the caller already has a copy on hand (init's
+// config-build phase fetches it once for mode/id derivation, moments before
+// this runs) — a second fetch bought no fresher data and risked observing a
+// different CR generation than the one derivation just used.
+func TestWorkflow_Initialize_ReusesPreFetchedCR(t *testing.T) {
```
adrian-januzi, line 91: Two things around this:
1. I think we should always regrab the 'initial' CR that is currently live at FSM `initialize` regardless if we already have a copy as there is a chance of drift between `kcp migration init` --> `kcp migration execute`.
2. Speaking of drift, this is a candidate for the reconciliation engine.

ifeeneyconfluent (approved):

PR #428, opened 2026-09-04
fix(deps): patch mkdocs-material and fast-uri dependabot alerts

adrian-januzi:
## What

Patches the two Dependabot alerts that touch a real execution path:

- **`mkdocs-material` 9.7.6 → 9.7.7** (`requirements-docs.txt`) — clears GHSA-xvg9-69gf-fjrf (DOM XSS in docs-site search suggestions via query parameter). The docs site is the only user-facing surface here.
- **`fast-uri` resolution `^3.1.5` → `^3.1.6`** (`cmd/ui/frontend`) — lockfile now resolves `fast-uri` to `3.1.7`, clearing GHSA-jqff-g426-hqxp, GHSA-f65p-4m7j-42xc, GHSA-fph4-wmhf-6fwf, GHSA-5jgf-p345-68v8. `fast-uri` reaches the client bundle via `ajv` ← `@rjsf/validator-ajv8`; the CVEs aren't reachable in kcp (ajv only parses schema URIs, never fetches), but the bump is a one-liner and clears the noise.

## Not fixed (dismissed separately)

The remaining open npm alerts — `browserslist`, `@humanfs/node`, `postcss-selector-parser`, `@hono/node-server`, `js-yaml`, `hono`, `ip-address`, `postcss` — are all transitive dependencies of the `shadcn` and `eslint` **devDependencies**. None are bundled into the Vite `dist/` that ships embedded in the Go binary and is served by Echo, so their code never executes in kcp's runtime. Dismissed on the dashboard as *vulnerable code not in execution path*.

## Test

Frontend `yarn install` regenerates the lockfile cleanly; `fast-uri` collapses to a single `3.1.7` entry across all consumers. No Go changes.

### DIFF
--- a/cmd/ui/frontend/package.json
+++ b/cmd/ui/frontend/package.json
@@ -51,7 +51,7 @@
   "resolutions": {
     "lodash": "^4.18.0",
     "lodash-es": "^4.18.0",
-    "fast-uri": "^3.1.5",
+    "fast-uri": "^3.1.6",
     "@rjsf/core/nanoid": "^6.0.0",
     "@rjsf/utils/nanoid": "^6.0.0",
     "**/postcss/nanoid": "^6.0.0"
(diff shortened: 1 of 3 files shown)


tjunderhill (approved):
LGTM!

PR #431, opened 2026-09-06
feat(scan): discover consumer groups with KIP-848 group types

pinaruyar:
## Summary

Adds **consumer-group discovery** to `kcp scan clusters` (MSK + Apache Kafka). A scan now captures each cluster's consumer groups into `kcp-state.json` and surfaces them in a new **Consumer Groups** tab in `kcp ui`.

Per group: id, **type** (`classic` / `consumer` / `share` / `streams`, via KIP-848 `ListGroups v5`), state, members (classic only), coordinator, and topics.

## Key design decisions

- **Isolated Kafka 3.8 client** for the group lookup (raw `ListGroups v5` for the KIP-848 type), leaving the existing 3.6-era scan clients untouched; `v5→v4` fallback for pre-3.8 brokers.
- **Type-dispatch:** the classic `DescribeGroups` API (all sarama implements) only describes `classic` groups correctly — for `consumer`/`share`/`streams` it returns a misleading `Dead`/empty answer, so those types keep their accurate `ListGroups v5` state and are flagged `detail_complete=false`. Full member detail for them needs the type-specific `*GroupDescribe` APIs (key 69 etc.), which sarama does not implement — a documented follow-up.
- **Schema bump v3→v4** (append-only shape freeze).
- **Degrade-not-abort:** a group-scan failure warns and records none, never sinking the cluster scan. Adds `kafka-cluster:DescribeGroup` to the scan IAM annotation.

## Validation

- **Unit tests** — mapping, type-dispatch (incl. the `Dead`→`Stable` fix), `v5→v4` fallback, resilience.
- **Integration matrix** (Docker, AK 3.7–4.2) exercising all four group types on the broker version where each is GA — wired into Semaphore as a **blocking** block (`integration: consumer group scan`).
- Validated live against a real **MSK 4.0** cluster over IAM.

## Not included (follow-ups)

- Type-specific describe (`ConsumerGroupDescribe` / API 69) for full `consumer`/`share`/`streams` member detail — needs a client that implements it (raw protocol or franz-go).
- Per-partition offsets / lag (deliberately deferred).

## Notes

- Second commit (`chore(deps)`) bumps `golang.org/x/crypto`, `echo/v4`, and `fast-uri` to patched versions — pre-existing HIGH/CRITICAL CVEs the Trivy scan flags, unrelated to this feature but required for green CI.
- **Review-only PR** per the hotfix process — do **not** merge in GitHub; merge locally with a squash.


adrian-januzi (approved):
lgtm

Image

Something nice is that kcp actually returns all consumer groups compared to a native KAfka admin `kafka-consumer-groups --list` call which wouldn't return the `connect` CGs.

pinaruyar:
you can also click on the groups individually to see the details in case you miss that @adrian-januzi

PR #433, opened 2026-09-08
feat(connect): metric-name overrides for Connect worker scans

pinaruyar:
## Description

Adds per-cluster metric-name/MBean overrides for Kafka Connect worker metrics (`prometheus.connect_metric_names`, `jolokia.connect_mbean_overrides`), mirroring the existing broker-metric override feature, for `kcp scan self-managed-connectors`. Also fixes a `GlobalPartitionCount` override bug on the existing broker-metric path and adds a configurable Prometheus request timeout, both reported against v0.9.0 while validating this feature.

---

## Changes Made

- [x] **Connect metric-name overrides**: `prometheus.connect_metric_names` / `jolokia.connect_mbean_overrides`, keyed on 10 canonical Connect metric labels, wired into the self-managed-connectors scanner, with validation, docs (`connect-metrics-collection.md`), and unit + Docker-based integration test coverage.
- [x] **Fix**: `GlobalPartitionCount`'s `prometheus.metric_names` override was nullified by an unconditionally-appended `{name="GlobalPartitionCount"}` PromQL discriminator — it's now only appended on the non-overridden (default) path, matching the fix suggested in the linked report.
- [x] **Feature**: configurable `prometheus.timeout` credentials field (Go duration string, e.g. `90s`), replacing the previously hardcoded 30s client timeout, wired into both the broker-metrics and Connect-metrics scan paths.
- [ ] No breaking changes — both new fields are optional/`omitempty` and zero-value-safe; existing credentials files behave identically.
- [x] Documentation updates (`credentials.md`, `metrics-collection.md`, `connect-metrics-collection.md`).

---

## Testing

- [x] New tests added/updated (unit tests for `internal/services/prometheus`, `internal/services/jmx`, `internal/types`, `internal/client`; Docker-based integration tests in `integration-tests/connect-scan` and `integration-tests/osk-scan`)
- [x] Manual testing completed (`go build ./...`, `go test ./...`, `make test-osk-scan`, `make test-kafka-connect` all green)
- [x] All tests pass

**Test Instructions:**

- `go test ./...` for unit coverage.
- `make test-kafka-connect` for the Connect metric-name-override integration suite (Docker Compose).
- `make test-osk-scan` for the broker-path `GlobalPartitionCount` override regression test (Docker Compose).

---

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed (per-task + final whole-branch review, all clean)
- [x] Documentation updated
- [x] No breaking changes


cmd/scan/clusters/cmd_scan_clusters.go
```diff
@@ -478,6 +478,9 @@ func collectPrometheusMetrics(ctx context.Context, clusterCreds types.OSKCluster
 		}
 		promOpts = append(promOpts, client.WithPrometheusTLS(caPool, clusterCreds.Prometheus.TLS.InsecureSkipVerify))
 	}
+	if clusterCreds.Prometheus.Timeout > 0 {
```
adrian-januzi: Just wondering if we should still have some sort of upper-bound for the timeout? I think previously it was just a hard cutoff at 30s which was obviously too short in hindsight.

I just don't know what would be an acceptable upper-bound - 15 minutes? 5 minutes? 1 hour?

internal/types/credentials_osk.go
```diff
@@ -380,7 +430,10 @@ func validateJolokiaConfig(jolokia *JolokiaConfig) error {
 			return fmt.Errorf("tls ca_cert file not found: %s", jolokia.TLS.CACert)
 		}
 	}
-	if err := validateMetricNameOverrides(jolokia.MBeanOverrides, "mbean_overrides"); err != nil {
+	if err := validateMetricNameOverrides(jolokia.MBeanOverrides, "mbean_overrides", validBrokerMetricLabels, ValidBrokerMetricLabels()); err != nil {
+		return err
+	}
+	if err := validateMetricNameOverrides(jolokia.ConnectMBeanOverrides, "connect_mbean_overrides", validConnectMetricLabels, ValidConnectMetricLabels()); err != nil {
```
adrian-januzi, line 445: nit: the 'unknown key' has no test.

internal/types/credentials_osk.go
```diff
@@ -343,6 +387,9 @@ func validatePrometheusConfig(prom *PrometheusConfig) error {
 	if prom.URL == "" {
 		return fmt.Errorf("url is required")
 	}
+	if prom.Timeout < 0 {
+		return fmt.Errorf("timeout must not be negative")
+	}
```
adrian-januzi, line 401: Just wondering if we should still have some sort of upper-bound for the timeout? I think previously it was just a hard cutoff at 30s which was obviously too short in hindsight.

I just don't know what would be an acceptable upper-bound - 15 minutes? 5 minutes? 1 hour?

---

Also, is there a default value for this or is it a required field in the credentials file? I think we should default to what it was previously (30s?) and keep that as a lower-bound?

adrian-januzi: I'm just thinking out loud here - do we have any tests around checking that we are honouring the set timeout?

On the flipside, if we do have a lower-bound @ 30s, it would mean we have at least 1 test that takes up to 30s to confirm the lower-bound :|

cmd/scan/clusters/cmd_scan_clusters.go
adrian-januzi, line 1: From what I can see, we have parsing tests for the field values but missing tests that actually use the `timeout` value provided and ensure it is enforced against the JMX broker which I think would live in the sister test file `cmd_scan_clusters_test.go`.

adrian-januzi (dismissed):

internal/types/credentials_osk.go
pinaruyar: yeah default was 30secs and now I have made 30s a lower bound as well but haven't put a upper bound yet, could stick to smth like 3-5min.

internal/types/credentials_osk.go
pinaruyar: now it does

cmd/scan/clusters/cmd_scan_clusters.go
pinaruyar: I have added the tests against Prometheus but one thing is that Prometheus path does not use the JMX/Jolokia client that's out of scope unf. But now we have tests that use timeout value

adrian-januzi: That'll do, ty

internal/types/credentials_osk.go
adrian-januzi: I think even if we just stick an upper-bound like 10-15 mins just to be safe? Though maybe others will have an opinion so we can ask in the standup tomorrow?

pinaruyar:
happy for me to merge or still reviewing @adrian-januzi ?

adrian-januzi:
@pinaruyar yes LGTM

pinaruyar:
Merged into `hotfix/v0.9.2` via local squash commit 9cb404adec15b4146bba3ae5a0e7b684dcc68bc0, following the kcp-release hotfix process (local squash + push; the GitHub merge button is intentionally not used, since Airlock mishandles PR merges to non-default branches). Closing.

PR #439, opened 2026-09-15
chore: speed up execute-tbm startup (observability + connection dedup)

adrian-januzi:
Speeds up `kcp migration execute-tbm` startup, and makes it possible to see what it's doing while it starts.

Three small changes:

1. **See what's happening during startup.** The underlying Kafka library was logging nothing, so slow connections were invisible. Its activity now lands in `kcp.log` (and on the console with `--verbose`), and the gateway fetch is timed.

2. **Stop connecting to each cluster twice.** The command was opening two separate connections to the source and to the target — one for reading offsets, one for listing topics. It now reuses a single connection per cluster.

3. **Stop opening an extra connection just to read the cluster ID.** That read now reuses a connection that's already open.

On the minikube test rig this brought startup down from ~60s to ~15–20s. Each change is behind its own commit and covered by tests.

## Change log

Reconstructed from the build sessions. This branch's own delta is the six commits below; the wider GitHub diff is inherited from the not-yet-merged base (see *Diff scope* at the bottom).

| Change | What & why | Evidence (from the build sessions) |
| --- | --- | --- |
| **Startup observability** (`43ab92b0`) | Bridge sarama's `Logger` **and** `DebugLogger` to `slog.Debug` (prefixed `sarama: `), installed once from the root command's `PersistentPreRun`; time the gateway-CR `Get()` and add `ms` to its log line. sarama stays at Debug, so broker-address detail is off the default Warn+ console but always in `kcp.log` (surfacing on `--verbose`). Adds an `auditedSaramaVersion` tripwire so a go.mod bump forces a re-audit. **Why:** sarama's loggers sat at their `io.Discard` defaults repo-wide, so every broker dial, metadata fetch, retry and coordinator lookup during startup was invisible — startup *looked* like a hang. | The work started from a report that `execute-tbm` **appeared to hang** at startup on silent Kafka dials. The gateway-fetch timing is covered test-first via an extracted `getGatewayYAML(ctx, dynamic.Interface, …)` seam. sarama v1.46.3 was checked to confirm it never logs SASL/SCRAM credentials. |
| **Reuse offset connections for topic listing** (`af5547a2`) | `NewKafkaAdminFromClient` wraps the already-dialed offset `sarama.Client` (an `ownsClient=false` flag makes its `Close()` a no-op, so the offset side keeps sole ownership); `migplan.WithSharedClients(src, tgt)` injects those into `Reconcile` so the topic listers ride the open connections. Dry-run and standalone `Reconcile` callers still dial fresh. **Why:** each cluster was dialed twice at startup — once as an offset client, once as a topic-listing admin — and each dial stalls ~5s reaching the advertised `svc.cluster.local` listener on the rig. | The new logging **named the culprit**: a 60s gap = ~10 "Connected to broker" events ~5s apart, with the gateway fetch negligible (`ms=10`). **Measured on the rig:** 4 client constructions → 2, ~10 broker connections → 6, startup **~60s → ~30s**. |
| **Read cluster ID over the shared connection** (`f21ad27b`) | A from-client admin reads the cluster id over the shared client's already-open controller connection; `getClusterIDFromBroker`'s fresh `NewBroker().Open()` is kept only for natively-dialed admins (scan/discover/report). Metadata-read logic factored into a shared `clusterIDFromBrokerConn`. **Why:** the reconcile engine's `ClusterID()` call still paid one fresh ~5s dial per cluster after the dedup. | By the time the engine asks for the cluster id it has already listed topics, so the controller connection is warm and no new dial is made. **Measured:** connections ~6 → 4, startup **~30s → ~15–20s** (the kcp-side floor). Tested behaviourally by counting `ApiVersions` handshakes (no `MockMetadataResponse` ClusterID setter exists), verified with a clean RED. |
| **Review fixes** (`17c85b4d`) | The main issue found while reviewing the three perf commits: `kcp migration execute` (the non-TBM path) *also* double-dialed → offset connections are now dialed in the command layer **before** the state-gated `Reconcile`, reused via `WithSharedClients`. Also: dial at `V3_6_0_0` to match dry-run's wire version; `offset.ClientOf` replacing a type assertion that could panic on a typed-nil `*Service`; idempotent `InstallSaramaLogging` (`sync.Once`); ownership derived from `client == nil` (dropped the drift-prone `ownsClient` field); concurrent source/target dials via `errgroup`; helpers de-duplicated into `internal/testsupport`; added shared-client-lister and go.mod-version-guard coverage. | Verified: `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues), `make test-go` (full suite) and `-race` on every touched package all pass. |
| **Trim comments / drop dead test** (`f24e0c60`) | Cut ~15 repeated multi-paragraph rationale comments to 1–3 lines across the touched files (the same "why" was restated in ~5 places); delete `share_clients_test.go`, now dead after `offset.ClientOf` (it only re-proved coverage already in `offset_test.go`). **Why:** the user flagged the PR added more than it removed for its size. | No behaviour change; net **−123 lines**. Build/vet/lint clean, full suite green. |
| **Merge reconciliation** (`56d73032`) | Merge `tbm-initialize-transition`, then `main` (which now carries the canonical TBM landing) into the branch, re-porting these deltas onto the unified `execute` command home. | One concrete fix: a `buildExecutorOpts` signature/caller mismatch between main (5-param) and this branch (4-param dedup version), resolved to the 4-param version. Full `make test-go` green after the first merge. |

## Not in this PR — the residual per-dial cost is environmental

After the three code changes, the remaining ~5s per fresh dial was root-caused to a **macOS `.local` IPv6/AAAA DNS stall**, not kcp: the release binary is pure-Go (`CGO_ENABLED=0`) and dials `"tcp"` (dual-stack), so it issues an AAAA lookup; `/etc/hosts` has only IPv4 A entries for `*.svc.cluster.local`, so AAAA misses, hits a VPN-range resolver that black-holes it, and Go times out at exactly 5.00s per dial. Fixed **rig-side** (two `::1` `/etc/hosts` entries, in the migration-tooling repo — not here), which took the rig from ~20s to ~1s. That is why the code-only speedup here is quoted as ~60s → ~15–20s rather than ~1s.

## Diff scope

This PR targets `main`, but its real base — `tbm-initialize-transition` (#434) — targets `migration-reconciliation-engine-proto` and is not yet on `main`. GitHub therefore shows the whole stacked diff. This PR's own change is only the six commits above (`43ab92b0..56d73032`).


adrian-januzi:
Superseded by #442.

This PR was stacked on #440; after #440 squash-merged to `main`, this branch carried all 126 of #440's commits as distinct history, so GitHub showed 100+ commits for a 16-file change. #442 is the same change rebuilt fresh off `main` as one commit with a byte-for-byte identical net diff. Closing in favour of it — review discussion here stays for reference.

PR #440, opened 2026-09-15
feat(migration): Topic Based Migration (TBM)

ifeeneyconfluent:
## Summary

Adds **Topic Based Migration (TBM)** — a way to move a Kafka cluster to
Confluent Platform/Cloud one batch of topics at a time, instead of all at
once. An operator fences a batch of topics on a Confluent Gateway dynamic
route, waits for cluster-link replication lag to drain, promotes the
batch's mirror topics, and switches production traffic over to the
destination — repeatable batch by batch until the whole migration is
complete. It complements the existing all-at-once (AAO) migration path
(`kcp migration execute` against a *static* gateway route), which this PR
also refactors onto the same underlying engine.

Driven by a single command:

```
kcp migration execute --manifest gateway-migration.yaml
```

which auto-registers a migration on first sight of its manifest's
`metadata.name`, resumes an in-flight one from its persisted state file on
every later run, and dispatches to the TBM (dynamic-route) or AAO
(static-route) FSM depending on the target route's resolved mode — no
separate command or flag needed to pick one.

### Architecture

- **`internal/services/migplan`** — a new, provider-agnostic reconciliation
  engine that is the single source of truth for "is this batch of topics
  safe to migrate, and what Gateway CR changes does it require": resolving
  a topic selector against the live source cluster, classifying each topic
  against source/target presence and cluster-link mirror state, checking
  run-level preconditions (route shape, coordination pinning, offset-sync,
  cluster identity), and rendering the fence/switchover artifacts to apply.
  It has two strategies — dynamic (TBM, per-topic routing) and static
  (AAO, whole-route redundant-auth switch) — sharing one refusal policy
  (any failed precondition or fail-fast topic emits no artifacts at all)
  and one topic-selection/classification pipeline.
- **TBM's FSM** (`internal/services/migration/tbm`, now unified with AAO's
  own FSM onto one shared `MigrationConfig`/`MigrationState` type) drives a
  batch through `initialize → wait_for_lags → fence → verify_fence →
  promote → switch`, with an `abort_fence` rollback if a producer is
  detected bypassing the fence during the monitoring window. Every
  transition is resumable: a killed or restarted run picks up from
  whatever the state file's `CurrentState` says, re-validating
  point-in-time facts (fence status, lag) rather than trusting them stale.
- **Command-layer unification**: `kcp migration init` and
  `kcp migration execute-tbm` are both retired. Registration, manifest
  validation, and drift detection are now handled entirely by
  `kcp migration execute` for both migration kinds — one command, one
  state file format, one `--dry-run` flag (validates everything, mutates
  nothing) for either kind of migration.

### Notable fixes found along the way

Building this surfaced and fixed several real, previously-latent bugs —
via code review, live e2e verification against a real Minikube/CFK
cluster, or both — including: a missing-lag guard for an already-migrated
zero-topic batch, a required Gateway CRD field the fence artifact never
set, a stale gateway-capability cache that could resume `switch` with the
wrong hot-reload assumption, an RBAC-permission-denial that was
incorrectly treated as a hard reconcile failure instead of an advisory
warning, and a config field (`Mode`) that resolved correctly but was never
persisted — meaning a resumed migration could silently dispatch to the
wrong engine. See #434 for the full, task-by-task engineering log (this
PR's `tbm-initialize-transition` branch, merged there against its interim
base `migration-reconciliation-engine-proto`) and the `migplan` engine's
own commit history for the reconciliation-engine build-out this rolls up.

## Test plan

- [x] `go test ./...` — full suite green (merged-result re-run included), `go vet ./...` / `gofmt -l .` / `golangci-lint run ./...` clean
- [x] Live e2e, real licensed Minikube/CFK cluster:
  - `make test-migration` (AAO, static-route) — 23/23 tests green
  - `make test-migration-tbm` (TBM, dynamic-route, hot reload) — full suite green across all FSM transitions (fence, verify_fence + rogue-producer detection + abort_fence rollback, promote, switch), including the halt-scenario precondition suite fixed on this branch
- [x] Every phase reviewed task-by-task and as a whole branch (subagent-driven-development); see #434 for each phase's findings and fix rounds
- [x] Manifest-drift, `--dry-run`, and resume/idempotency behavior covered by both unit and live e2e tests for both migration kinds


cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -317,24 +449,23 @@ func detectDrift(g *manifest.GatewayMigration, config *migration.MigrationConfig
...
+		}
+		if entry.TargetStreamingDomain != config.TargetDomain {
+			topicGroupChanges = append(topicGroupChanges, "target domain")
+		}
+		// A match-all topicGroup selection now resolves via migplan's Explode
+		// against source topics, exactly like an explicit list — no special
+		// "whatever the cluster link mirrors" case remains, so drift compares
+		// the manifest's declared topics against the snapshot the same way for
+		// both topics and topicPatterns entries.
+		if topics := entry.Topics; topics != nil {
```
adrian-januzi, line 476: **Drift check ignores `spec.topicGroup[0].topicPatterns`.**

`detectDrift` diffs `entry.Route`, `entry.TargetStreamingDomain`, and `entry.Topics`, but never `entry.TopicPatterns`. For a pattern-selected topic group `entry.Topics` is `nil`, so this whole block is skipped and no drift is reported.

Scenario: a manifest selects `topicPatterns: ["foo.*"]`; the first run resolves and snapshots `config.Topics=[foo1,foo2]`. An operator edits it to `["bar.*"]` and re-runs. `checkSpecDrift`/`detectDrift` runs but reads nothing pattern-related, so it reports no drift — and because reconcile only re-runs at `StateUninitialized`, the run silently proceeds against the stale `[foo1,foo2]` snapshot, violating the "topology must not change once registered" guarantee.

The comment just above (lines 463–467) claims patterns are compared "the same way" as topics, which isn't true. A `TopicPatterns` comparison needs to be added here.

cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -120,18 +200,48 @@ func runMigrationExecute(cmd *cobra.Command, args []string) error {
...
+		if kerr != nil {
+			return kerr
+		}
+		fresh := buildFreshMigrationConfig(g, id, kubeConfigPathResolved)
+		config = &fresh
+		state.UpsertMigration(*config)
+		if err := state.WriteToFile(migrationStateFile); err != nil {
+			return fmt.Errorf("failed to write migration state file: %w", err)
+		}
+	} else if err := checkSpecDrift(g, config); err != nil {
```
adrian-januzi, line 252: **Removing the `checkStateFilePredatesSwitchover` pre-flight guard lets a legacy/incomplete state file fail deep inside Fence/Switch instead of before any cluster contact.**

The guard used to run right here (before `checkSpecDrift`). A `migration-state.json` missing the new `gateway_yaml`/`fence_yaml`/`switchover_yaml`/`mode` fields (written by the removed `init`, or a registration whose `Initialize` never completed) resumes at a state past `StateUninitialized`. So `mode := config.Mode` is `""`, reconcile is skipped (it only runs at `StateUninitialized`), and it dispatches to the AAO/default branch.

`ensureGatewayCapability` → `deriveFencedCRYAML`/`deriveSwitchedCRYAML` then call `parseGatewayYAML` on an empty `config.GatewayYAML` and splice an empty fragment, erroring with e.g. "fence fragment has no top-level fence key" — partway through `SwitchGateway`, after the gateway may already be fenced from the prior run — rather than the pre-flight refusal this guard used to provide.

internal/services/gateway/dynamic_route.go
```diff
@@ -0,0 +1,81 @@
...
+	found := false
+	for _, raw := range routes {
+		route, ok := raw.(map[string]any)
+		if !ok {
+			continue
+		}
+		if name, _ := stringField(route, "name"); name != routeName {
+			continue
+		}
+		found = true
```
adrian-januzi, line 65: **`ReplaceRouteRulesObj` replaces rules on _every_ name-matching route (the loop has no `break`/`return`), while the static path's `replaceRouteField` returns on the first match.**

If a hand-edited/malformed gateway CR has two `spec.routes[]` entries with the same `name`, this dynamic fence/switch path overwrites rules on both, whereas `replaceRouteField` in `static_route.go` mutates only the first. The two splice helpers disagree on the same CR shape. Either `break` after the first match to match the single-route contract, or make both paths reject a duplicate-name CR.

internal/services/migplan/reconcile/reconcile.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Build both artifacts from one pristine copy of the operator's rules, so the
+	// fence and switchover derive independently from the same baseline.
+	base, _ := ParseRules(gw.Route.Rules)
+	fence := base.Clone()
+	fence.PrependFence(migratable)
+	switchover := base.Clone()
+	switchover.PrependCondition(migratable, view.TargetDomain)
+
+	fenceBytes, _ := fence.Serialize()
```
adrian-januzi: **`Serialize()` errors are discarded, so an empty artifact slips past the size guard and surfaces as a confusing derive-time error.**

`fenceBytes, _ := fence.Serialize()` / `switchBytes, _ := switchover.Serialize()` drop the error. On a marshal failure the returned bytes are empty (len 0), which passes the `> MaxRulesBytes` check just below and gets stored as `Artifacts.FenceRules`/`SwitchoverRules`. At execute time `gateway.ReplaceRouteRulesObj` then parses the empty YAML and returns "rules YAML has no top-level rules key" mid-fence, instead of reconcile refusing the plan up front. Handle the error and append a `fail(...)` precondition rather than swallowing it.

internal/services/migplan/reconcile/staticartifacts.go
```diff
@@ -0,0 +1,37 @@
+package reconcile
+
+import "github.com/goccy/go-yaml"
+
+// staticFenceScope and staticFenceErrorCode are the fence parameters this
+// migration injects — the same values gateway.FenceRoutes hardcodes today
+// (see internal/services/gateway/fence.go). Duplicated here per the migplan
+// static-route-strategy design doc's decision 13 (no new dependency from
```
adrian-januzi: **Comment references an external, un-committed design doc.**

"per the migplan static-route-strategy design doc's decision 13" points at an artefact that doesn't live in the repo, so it's unresolvable to any reader once the plan is discarded. The same pattern appears across several new files — `staticpreconditions.go` ("decision 10"/"decision 4"), `reconcile.go` ("decision 6/7", "decision 8"), `gatewayfile.go` ("decision 14"), `tbm_dependencies.go`/`tbm_executor.go` ("Ported from … deleted in Task 1", "this plan's Task 2, Step 3"), and `cmd_migration_execute.go` ("this plan's Global Constraints").

Keep the rationale (why the values are duplicated here rather than imported) but drop the doc/decision/task pointers.

internal/services/migration/tbm/gateway.go
```diff
@@ -0,0 +1,326 @@
+package tbm
```
adrian-januzi, line 1: **This package re-implements ~300 lines of gateway-verification and offset-sweep logic verbatim from `internal/services/migration`.**

`gateway.go`, and `workflow.go`'s `fetchSourceAndDestinationOffsets`/`detectUnroutedProducers`/`formatLag64`/`formatElapsed`, plus the apply/wait/verify helpers, are all documented as "Mirrors migration.X verbatim". A correctness fix (a race, a timeout, an emoji/log change) applied to one package silently leaves the other stale — the same split-path regression hazard we already have with connector-metrics. Consider extracting the shared verification/offset primitives into a helper both packages call, rather than freezing two hand-synced copies.

internal/services/migplan/reconcile/staticpreconditions.go
```diff
@@ -0,0 +1,227 @@
...
+	ok := true
+	for _, r := range res {
+		if !r.OK {
+			ok = false
+		}
+	}
+	return res, view, ok
+}
+
+func joinNames(names []string) string {
```
adrian-januzi: **`joinNames` reimplements `strings.Join(names, ", ")`.**

Eight lines duplicating the stdlib with no behavioral difference — the package already deals in the same `[]string`. Replace the call sites with `strings.Join(names, ", ")` and drop this helper.

internal/services/gateway/dynamic_route.go
adrian-januzi: I don't think this is possible because the gateway YAML will be rejected and never actually running with this scenario of 'two routes with the same name'. However, I wonder if it is possible for us to pull down the live YAML of a crashlooped gateway? Should we check its status before doing anything?

cmd/migration/execute/cmd_migration_execute.go
adrian-januzi: I think we can ignore this for now as a low-prio. However, should we keep legacy `migration-state.json` files in mind for the future.

internal/services/migplan/engine.go
```diff
@@ -0,0 +1,83 @@
...
+type ReconciliationEngine struct {
+	gateway GatewayConfigSource
+	source  TopicLister
+	target  TopicLister
+	link    LinkStatusProvider
+	secrets SecretExistenceChecker
+}
+
+func NewReconciliationEngine(gateway GatewayConfigSource, source, target TopicLister, link LinkStatusProvider, secrets SecretExistenceChecker) *ReconciliationEngine {
+	return &ReconciliationEngine{gateway: gateway, source: source, target: target, link: link, secrets: secrets}
```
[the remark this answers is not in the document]
ifeeneyconfluent: Fixed in 85367ecc: `newLiveEngineFor` / the two direct `NewReconciliationEngine` call sites now pass a `fakeSecretChecker` test stub as the 5th arg, so `test-migplan` compiles again. Also wired `make test-migplan` into `.semaphore/semaphore.yml` (it was not running in CI at all before, which is why this slipped through), and added static-route coverage to the suite while I was in there — verified live, 14/14 green.

internal/services/gateway/static_route.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	for _, raw := range routes {
+		route, ok := raw.(map[string]any)
+		if !ok {
+			continue
+		}
+		if name, _ := stringField(route, "name"); name != routeName {
+			continue
+		}
+		route[key] = value
```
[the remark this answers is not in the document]
ifeeneyconfluent: Fixed in 1a6e0096 — but not where you'd expect: rather than restoring this check inside `ReplaceRouteFenceObj` (apply-time), I moved responsibility into the reconcile engine itself, as a new `CheckStaticPreconditions` check — `"route is not already fenced"`. Consumers of a non-refused `Result`/`Plan` can now trust the plan already ruled out double-fencing, so no separate apply-time guard is needed here at all. Covered by two new unit tests plus a live e2e permutation (`gateway-static-already-fenced.yaml`).

internal/services/migplan/reconcile/reconcile.go
ifeeneyconfluent: Fixed in c0b2de90 — but writing a real test for it surfaced a worse, identical bug one step earlier: `RulesTree.Clone()` was *also* discarding its own marshal/unmarshal errors, silently handing back a corrupt (nil-root) copy. Since `PrependFence`/`PrependCondition` write into that root unconditionally, a failed `Clone()` would panic on the very next line — meaning your original `Serialize()` fix alone could never actually be reached/exercised by any real input. Fixed both: `Clone()` now returns `(*RulesTree, error)`, and `reconcileDynamic` checks all four calls (2x Clone, 2x Serialize) via the normal refusal path. Covered by a primitive-level test plus a true end-to-end test through `Reconcile()`.

internal/services/migplan/engine.go
```diff
@@ -0,0 +1,83 @@
...
+
+	plan := reconcile.Reconcile(in, gw, src, tgt, link.Mirrors, link.OffsetSyncEnabled, ids, missingSecrets)
+	// A permission denial is a skip, not a precondition failure — surfaced as
+	// a warning (never blocking) rather than folded into missingSecrets,
+	// which would otherwise read as "these specific secrets don't exist"
+	// when the truth is "we couldn't check at all". See
+	// SecretExistenceChecker's own doc comment for why this distinction
+	// matters.
+	if secretCheckSkipped != "" {
+		plan.Report.Warnings = append(plan.Report.Warnings, secretCheckSkipped)
```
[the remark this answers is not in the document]
ifeeneyconfluent: Fixed in 5074c5f0.

Added a `Skipped bool` on `PreconditionResult` (paired with `OK: true`) and threaded the skip reason through `Reconcile` → `reconcileStatic` → `CheckStaticPreconditions`. The "staged auth secrets exist" precondition now records `Skipped: true` with the skip reason as its detail, and `RenderReport` prints skipped preconditions as a yellow `⚠` rather than the green `✓` a real pass gets — so the rendered report no longer claims a check happened when it didn't.

The existing `Report.Warnings` append in `engine.go` (which a prior incident-driven regression test depends on) is left untouched — the warning and the new `Skipped` precondition are now two consistent views of the same event instead of contradicting each other.

Test coverage: `TestStaticPreconditionsSecretCheckSkipped`, `TestReconcileStaticSecretCheckSkippedDoesNotRefuse`, `TestRenderReportSkippedPrecondition`, plus strengthened assertions on the pre-existing `TestReconciliationEngine_Run_StaticMode_SecretCheckSkippedIsWarningNotRefusal`.

cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -102,16 +121,77 @@ func NewMigrationExecuteCmd() *cobra.Command {
...
 	}
 
+	// --dry-run stops here: the reconcile step is self-contained (it opens its
+	// own live Gateway CR + source/target/cluster-link reads directly from the
+	// manifest) and renders its own report to the command's writer. Nothing
+	// past this point — the migration state file, config resolution, drift
+	// checking, any FSM transition — runs under --dry-run. Mirrors
+	// execute-tbm's identical --dry-run branch.
+	if dryRun {
+		res, err := migplan.Reconcile(cmd.Context(), g, migplan.WithOutput(cmd.OutOrStdout()))
```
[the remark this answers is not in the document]
ifeeneyconfluent: Fixed in 31a98e4e.

Moved `applyPolicyOverrides` + `DefaultPolicies.Validate()` to run unconditionally before the `--dry-run` check, so both paths validate the effective policy identically. Both are pure in-memory operations, so there's no behavioral cost to running them earlier — dry-run still stops before touching the state file, drift checking, or any FSM transition.

Added `TestExecute_DryRun_ValidatesPolicyOverrides`: `--dry-run --lag-threshold=-1` now fails with the policy-validation error, and returns fast (0.01s) confirming it never reaches the reconcile call.

cmd/migration/execute/cmd_migration_execute.go
ifeeneyconfluent: Fixed in 8254f7df.

Added a `TopicPatterns []string` snapshot on `MigrationConfig`, captured at registration from the manifest's declared `topicPatterns` (mirroring how `Route`/`TargetDomain` are already snapshotted there). `detectDrift` now compares `entry.TopicPatterns` against that snapshot declared-to-declared, independent of the resolved `Topics` snapshot — so an edited pattern is caught as drift even if it happened to expand to the same topics today.

Test coverage: `TestDrift_DetectsChangedTopicPatterns` (regression test for this exact scenario), `TestMigrationConfig_EveryFieldClassifiedForDrift` updated for the new field, and `TestBuildFreshMigrationConfig_PopulatesManifestFields` extended to assert the snapshot is captured at registration.

internal/services/gateway/dynamic_route.go
ifeeneyconfluent: Fixed in 746d6acf — added a `break` after the first name-matching route, so this now matches `replaceRouteField`'s (static_route.go) single-route, first-match contract for the identical CR shape.

Added `TestReplaceRouteRulesObj_OnlyFirstDuplicateNamedRouteChanges`: a fixture with two same-named routes, asserting only the first is patched.

Re your follow-up question about a crash-looped gateway and checking its status first — replying on that thread separately.

ifeeneyconfluent: Good question, but I'd treat it as a separate architectural discussion rather than something to fold into this PR — checking a gateway's rollout/crash-loop status before pulling its live CR would be a new precondition on top of what migplan currently does, and touches the gateway-capability resolution path more broadly, not just this fence/switch splice helper. Happy to file a follow-up if you'd like to pursue it, but leaving it out of scope here. I did add the `break` fix on the sibling thread regardless, for consistency with the static path.

internal/services/migplan/reconcile/staticpreconditions.go
ifeeneyconfluent: Fixed in 46b4dcb7 — replaced the call site with strings.Join(missingSecrets, ", ") and dropped joinNames.

internal/services/migplan/reconcile/staticartifacts.go
ifeeneyconfluent: Fixed in 46112dae.

Cleaned up the doc/decision/task pointers in the files you named (staticartifacts.go, staticpreconditions.go, reconcile.go, gatewayfile.go, tbm_dependencies.go, tbm_executor.go, cmd_migration_execute.go), keeping each comment's underlying rationale — plus one you didn't call out but has the identical pattern: secrets.go:16 ("see the migplan static-route-strategy design doc, decision 4").

Left the same pattern in a handful of test-file comments (decision numbers in gatewayfile_test.go, gatewaylive_test.go, reconcile_test.go, preconditions_test.go, cmd_migration_lagcheck_test.go, manifest_fixture_test.go, gateway_test.go) out of scope for now — lower-visibility than shipped comments, happy to clean those too if you'd like.

adrian-januzi (approved):
my claude is happy with your claude

internal/services/migration/tbm/gateway.go
ifeeneyconfluent: Fixed in efde314c — went with extracting the shared logic rather than leaving the two copies.

Moved the confirmed byte-for-byte duplicates into two neutral service packages both `migration` and `tbm` already depend on:
- `gateway.TransitionVerifier` (`transition_verifier.go`, `format.go`): the apply/wait/verify mechanism (`ApplyCR`, `ApplyConfigIDOnly`, `WaitForAccepted`, `WaitForConfigApplied`, `VerifyTransition`, `ResolveCapability`, `VerifyHotReloadCapability`) plus `FormatElapsed`/`FormatLag64`.
- `offset.DetectUnroutedProducers` / `offset.FetchSourceAndDestinationOffsets` (`unrouted_producers.go`): the two-snapshot check and the concurrent offset sweep. `ErrUnroutedProducers` is now canonically `offset.ErrUnroutedProducers`, with each package's existing sentinel re-exported as an alias so every existing `errors.Is` call site kept working unchanged.

Each package's genuinely different behavior stayed in its own thin wrapper around the shared mechanism — e.g. migration's `GatewayVerificationMode`/`GatewayHotReloadEnabled` persistence and mode-change warning, its legacy-zero-value port fallback, and TBM's per-pod-verification success message that migration never had. `tbm/state.go`'s package doc (which stated no code was shared between the two packages by design) is also updated — that separation was real when TBM was first built, but these are genuinely shared service modules now, not either package's own workflow logic.

Full repo build/vet/lint/test all clean after the extraction.

cmd/migration/execute/cmd_migration_execute.go
```diff
@@ -141,33 +251,51 @@ func runMigrationExecute(cmd *cobra.Command, args []string) error {
...
 	slog.Info("executing migration with effective policy", effectivePolicyLogArgs(id, config.CurrentState, g.Spec.DefaultPolicies)...)
 
-	if err := checkSpecDrift(g, config); err != nil {
-		return err
+	// migplan.Reconcile is called here — not by either FSM's Initialize itself
+	// — ONLY when resuming a migration still at StateUninitialized. Every other
+	// starting state skips this: the mode was already resolved and persisted by
+	// a prior run's Initialize step (see config.Mode, driftExempt because it
+	// "reflects the cluster's shape" — resolved once, live, and never
+	// re-derived on resume for either mode), so a live Reconcile call here
```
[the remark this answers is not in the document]
ifeeneyconfluent: Treating this as N/A — there are no in-flight migrations on main today whose state file predates Mode/Route/TargetDomain/GatewayYAML, so there's nothing on disk that would actually hit this path. Not adding a load-time upcast for a scenario with no real data to migrate. If that ever changes (e.g. a long-lived state file from before this PR is still in use somewhere), happy to revisit.

cmd/migration/execute/cmd_migration_execute.go
ifeeneyconfluent: Agreed, leaving this as-is for now — no code change. Same underlying category as the other legacy-state-file thread (Copilot's, on Mode/Route/TargetDomain/GatewayYAML): no real state file on main today is old enough to hit either path. Worth keeping in mind together if we ever need a load-time upcast pass for pre-this-PR state files.

adrian-januzi (approved):

PR #443, opened 2026-09-17
feat(migration): mutate Gateway CR via targeted JSON Patch (drop server-side apply)

adrian-januzi:
## What & why

`kcp migration execute` mutated the running Gateway CR by re-marshalling the **entire** CR from migplan's captured snapshot, splicing in the one route field that changed, and writing it back with a **full-object server-side apply** (`Force: true`). Two problems: (1) a stale snapshot clobbers fields kcp never intended to touch (AAO unfence re-applied the whole snapshot verbatim), and (2) `Force` seizes field ownership from the CFK operator and drags in SSA prune-on-omit semantics (the reason for the second disjoint field manager).

This PR makes the write **surgical**: each cutover step now issues a targeted **RFC 6902 JSON Patch** touching only the one route field it changes (plus `spec.configId`), guarded by a `test` op on the route's name at its resolved index, with **no field manager and no `Force`**. Verified live on minikube (CFK operator v0.1906.0) that a raw JSON Patch to `spec.routes` is not reverted by the operator and the `test`-op index guard is evaluated server-side.

## What changed

- **New pure primitives** (`internal/services/gateway`): `jsonpatch.go` (op-builder — `test` + `add`/`replace` + gated `add /spec/configId`, no `remove` op ever) and `route_patch.go` (`FragmentValue`/`RouteObject` extractors replacing the graft-onto-CR half of the old splicers).
- **New write path** on `K8sService`: `PatchGatewayRoute` / `PatchGatewayConfigID` read the live CR, resolve the route index, and `Patch(..., types.JSONPatchType, ...)` with no ownership. `TransitionVerifier` gains `PatchCR` / `PatchConfigIDOnly` (configId capability-gate + read-back contract identical to the old `ApplyCR`/`ApplyConfigIDOnly`).
- **AAO & TBM FSMs** cut over to derive a `RoutePatch` (fence/switch as a single field set; unfence as a whole-route `replace` restoring the captured route) and write via `PatchCR`. The **hot-reload probe** uses `PatchConfigIDOnly`.
- **SSA path deleted**: `ApplyGatewayYAML`/`ApplyGatewayConfigID` (+ inner funcs), both field-manager constants, the `prepare*`/`injectConfigID` helpers, and `ApplyCR`/`ApplyConfigIDOnly` — ~825 lines removed. The write path is now exclusively JSON Patch.
- **e2e**: gateway-contract tests moved to the patch path; SSA-ownership/full-CR-apply-premise tests removed. The fence-dataplane test was **converted** (not deleted) — it asserts real data-plane fencing. Two `DetectCapability`-only refusal tests were **kept** (they never drove SSA apply).

## Not in scope (by design)

- **No CRD change** — we do not rely on `x-kubernetes-list-map-keys` on `spec.routes` (that would enable SSA partial-apply-by-name but keeps ownership).
- **No change to migplan's reconcile engine** or the fragments it produces, and **no change to what each step semantically does**.
- **No `types.State` shape change** — `RoutePatch` is transient; the state schema snapshot is untouched.
- The route-splicers (`static_route.go`/`dynamic_route.go`) are **retained**: they still serve the unchanged **capability-detection** read path (`ResolveGatewayCapability`), which feeds full-CR bytes to hot-reload detection — a spec non-goal to change.

## Testing

- `go test ./...` — green (incl. `internal/types` + `internal/state/migrate` schema-snapshot guards; no persisted-shape change).
- `go build -tags e2e ./integration-tests/...` — clean; gateway-contract e2e suites migrated to the patch path.
- `golangci-lint` — clean on `internal/services/gateway/...` and `internal/services/migration/...`.

## Follow-ups (tracked, out of scope here)

- Re-evaluate `capability.go`'s hot-reload-mismatch refusal: it is premised on whole-CR SSA pruning of an omitted, already-owned `spec.hotReload`, which the route-scoped patch path can no longer trigger.
- Optionally give the `test`-op failure a distinct "route index moved" error (today it surfaces as a generic patch rejection; the FSM's resume-on-rerun re-reads the index).
- One residual `errors.New` string on `ErrApplyUnverified` still reads "applied" (cosmetic).
- The design notes a possible patch-path convergence e2e test to re-add the CFK-contract coverage the SSA-premise deletions dropped.


internal/services/gateway/gateway.go
```diff
@@ -169,120 +153,109 @@ func (s *K8sService) CheckPermissions(ctx context.Context, verb, resource, group
...
-	gatewayGVR := schema.GroupVersionResource{
-		Group:    GatewayGroup,
-		Version:  GatewayVersion,
-		Resource: GatewayResourcePlural,
+// patchGatewayRoute is the inner orchestration used by PatchGatewayRoute.
+// Split from the method so unit tests can inject a fake dynamic client.
+func patchGatewayRoute(ctx context.Context, dynamicClient dynamic.Interface, namespace, gatewayName string, rp RoutePatch, configID string) (string, error) {
+	gatewayGVR := schema.GroupVersionResource{Group: GatewayGroup, Version: GatewayVersion, Resource: GatewayResourcePlural}
+
+	live, err := dynamicClient.Resource(gatewayGVR).Namespace(namespace).Get(ctx, gatewayName, metav1.GetOptions{})
```
adrian-januzi, line 191: This live GET runs on every fence/switchover/unfence write, solely to resolve `rp.RouteName` to its `spec.routes` index (the `routes` slice then feeds `buildRoutePatchOps` below).

An alternative would be to have the reconciliation engine return the index alongside the YAML snippets.

I know we discussed this is drift and should just accept it for now but thought there was no harm in just implementing it now as a temp measure 🤷‍♂️

internal/services/gateway/route_patch.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/goccy/go-yaml"
+)
+
+// FragmentValue parses a single-key migplan fragment ({fence: ...},
+// {streamingDomain: ...}, {rules: ...}) and returns the value at key — the
+// fragment-extraction half of the old ReplaceRoute*Obj splicers, now that the
+// graft-onto-CR half is a JSON Patch built by the gateway service.
+func FragmentValue(fragment []byte, key string) (any, error) {
+	var m map[string]any
+	if err := yaml.Unmarshal(fragment, &m); err != nil {
```
adrian-januzi, line 15: **This is where the reconciliation engine YAML fragments are turned into the value that ends up in the JSON Patch(s).**

**Before:** the migration code used the `ReplaceRoute*Obj` splicers, which did two jobs at once — pull the value out of the fragment *and* graft it onto the full captured Gateway CR — after which the whole spliced CR was written back with a server-side apply.

**After:** that work is now split. `FragmentValue` is only the extraction half: it `yaml.Unmarshal`s a single-key fragment (`{fence: ...}`, `{streamingDomain: ...}`, `{rules: ...}`) into a Go `any` and returns the value under `key`. The graft-onto-CR half is now a narrow RFC 6902 JSON Patch built by the gateway service (`buildRoutePatchOps`), and this Go value rides along as an op's `value`.

So the YAML→JSON conversion is two-stage: **YAML→Go here** (`yaml.Unmarshal`), then **Go→JSON** where the ops are serialized (`gateway.go`, `json.Marshal(ops)`). See the notes on those spots.

internal/services/gateway/gateway.go
```diff
@@ -169,120 +153,119 @@ func (s *K8sService) CheckPermissions(ctx context.Context, verb, resource, group
...
+	if !found {
+		return "", fmt.Errorf("gateway %q has no spec.routes", gatewayName)
 	}
 
-	obj, err := prepareGatewayApply(yamlData, namespace, gatewayName, configID)
+	ops, err := buildRoutePatchOps(routes, rp, configID)
 	if err != nil {
 		return "", err
 	}
+	data, err := json.Marshal(ops)
```
adrian-januzi, line 207: **Second/final stage of the conversion.** The ops list — whose `value` fields carry the Go values parsed from YAML by `FragmentValue` / `RouteObject` — is serialized here into the RFC 6902 JSON Patch body sent to the API server. This is the point where the YAML-sourced data actually becomes JSON.

internal/services/gateway/route_patch.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	return v, nil
+}
+
+// RouteObject parses the captured gateway CR (config.GatewayYAML) and returns
+// the route element named routeName from spec.routes. Unfence uses it to
+// restore the route to its captured state via a whole-route replace.
+func RouteObject(gatewayYAML []byte, routeName string) (map[string]any, error) {
+	var obj map[string]any
+	if err := yaml.Unmarshal(gatewayYAML, &obj); err != nil {
```
adrian-januzi, line 33: **Same YAML→Go parse as `FragmentValue`, but for unfence's whole-route replace.** Parses the captured Gateway CR YAML and returns the named `spec.routes` element, which becomes the `replace` op's `value` — then JSON-encoded downstream at `gateway.go` `json.Marshal(ops)`.

internal/services/gateway/route_patch.go
adrian-januzi: Concrete example — the fence path (`config.FenceYAML`, built by `BuildFenceFragment`):

**YAML fragment (input):**
```yaml
fence:
  scope: ALL
  errorCode: BROKER_NOT_AVAILABLE
```

`FragmentValue(fragment, "fence")` unwraps the top-level key and returns the Go value:
```go
map[string]any{"scope": "ALL", "errorCode": "BROKER_NOT_AVAILABLE"}
```

The gateway service resolves the route to its `spec.routes` index and grafts that value onto a JSON Patch op (`buildRoutePatchOps`). After `json.Marshal(ops)` at `gateway.go:207`, the patch body sent to the API server is (route at index 0, first fence write so no field `test` op, no `configId`):
```json
[
  {"op":"test","path":"/spec/routes/0/name","value":"migration-route"},
  {"op":"add","path":"/spec/routes/0/fence","value":{"errorCode":"BROKER_NOT_AVAILABLE","scope":"ALL"}}
]
```
(`encoding/json` sorts map keys, so `errorCode` lands before `scope`.)

The switchover path is the same shape with `streamingDomain: {name, bootstrapServerId}`, and TBM's `rules` fragment likewise — only the top-level key and value differ.

internal/services/gateway/gateway.go
```diff
@@ -169,120 +153,119 @@ func (s *K8sService) CheckPermissions(ctx context.Context, verb, resource, group
...
-	slog.Debug("🔍 applying gateway CR (server-side apply)", "namespace", namespace, "gateway", gatewayName, "bytes", len(yamlData), "configId", configID)
+	slog.Debug("🔍 patching gateway CR (JSON patch)", "namespace", namespace, "gateway", gatewayName, "route", rp.RouteName, "field", rp.Field, "configId", configID)
 	start := time.Now()
-	applied, err := dynamicClient.Resource(gatewayGVR).Namespace(namespace).
-		Apply(ctx, gatewayName, obj, metav1.ApplyOptions{
-			FieldManager: migrationFieldManager,
-			Force:        true,
-		})
+	patched, err := dynamicClient.Resource(gatewayGVR).Namespace(namespace).
+		Patch(ctx, gatewayName, types.JSONPatchType, data, metav1.PatchOptions{})
```
adrian-januzi, line 215: **Why `JSONPatchType` (RFC 6902) and not JSON Merge Patch (RFC 7386):**

- `spec.routes` is an **array**. JSON Merge Patch replaces arrays wholesale, so touching one route would mean resending the entire array — no better than the full-object apply this change replaces, and it re-clobbers sibling routes. Ruled out.
- Strategic merge patch / SSA partial-apply-by-name would need `x-kubernetes-list-map-keys: [name]` on the CRD's `spec.routes`; the live CRD declares it as a plain `array` with no `x-kubernetes-list-type`/`x-kubernetes-list-map-keys`, and adding that annotation is a non-goal (no CRD change, and it re-introduces the field ownership we're trying to drop).
- JSON Patch addresses `/spec/routes/<index>/<key>` directly, replaces a whole value atomically (matching the existing "whole-subtree replacement, not a merge" guarantee for `rules`), needs no CRD support, and a `test` op guards the resolved index against a concurrent reorder.

So JSON Patch by index is the only flavour that targets one route surgically without a CRD change — merge patch is unusable for the routes array.

internal/services/gateway/gateway.go
ifeeneyconfluent: This get me wondering two things:

1. Should we just go ahead and do a drift check since we are getting the YAML again fresh just before everywrite.

2. What are the failure paths now?  Let's say the Gateway Did change and routeName can't be indexed.  What happens then?

adrian-januzi: > What are the failure paths now? Let's say the Gateway Did change and routeName can't be indexed. What happens then?

There are two failure cases called 'test' cases ([JSON Patch term](https://datatracker.ietf.org/doc/html/rfc6902#section-4.6)): 
- `test /spec/routes/2/name == "migration-route"` --> did the route move or get renamed? (index not resolvable)
- `test /spec/routes/2/fence == <old fence value>` --> did the field I'm about to patch change? (another route taking the index of the route being actioned on as per the migration)

So when patching, the flow is as follows:

1. we get the path that we are going to update - `/spec/route/2/fence` called `migration-route`.
2. before we apply the patch, make run the two tests mentioned above.
   2a. pass - continue.
   2b. fail - return to previous state.
3. patch the route.

Claude provided this snippet which I think is helpful to visualise where the `"op": "test"` asserts that this path equals 'X' and if any of the tests return false, the server rejects the patch:
```json
[
  {"op": "test",    "path": "/spec/routes/2/name",  "value": "migration-route"},
  {"op": "test",    "path": "/spec/routes/2/fence", "value": {"scope": "ALL", "errorCode": "BROKER_NOT_AVAILABLE"}},
  {"op": "add",     "path": "/spec/routes/2/fence", "value": {"scope": "ALL", "errorCode": "BROKER_NOT_AVAILABLE"}}
]
```

---

wrt. to the drift-check question - do you mean enhancing the scope from what is being above? If so, maybe? Though it's probably not _as_ pressing as other things on the TODO (imo).

ifeeneyconfluent: Didn't pick up on the `"op": "test"` so I think think addresses my  drift-check point

ifeeneyconfluent (approved):

internal/services/gateway/gateway.go
adrian-januzi: Following up with the above, there is one case where we might leave the Gateway in a fenced state...

If we fence the Gateway and before/during promotion something goes wrong **and** the route's name is updated between fence and switchover; we have no way of recovering without Omar intervention. They'll need to go into the `gateway-migration.yaml` manifest and update the `spec.route.name` field to the latest name.

SUPER EDGE CASE but worth flagging (even just to scare you).

PR #445, opened 2026-09-18
refactor(manifest): move gateway migration topicGroup/route under spec.route

ifeeneyconfluent:
## Summary
- Introduces `spec.route` on the `GatewayMigration` manifest: `topicGroup` and `targetStreamingDomain` move under it, and `topicGroup[].route` is renamed/moved to `spec.route.name`. `topicGroup[].topics`/`topicPatterns` are unchanged.
- Updates the Go structs (`internal/manifest/gateway.go`), the schema generator, and regenerates `gatewaymigration.schema.json`.
- Updates every call site (`migplan`, `migration execute` drift check), test fixtures, integration-test manifests, all 16 TBM batch templates, and the reference docs that encoded the old shape.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...` and `go vet -tags e2e ./...`
- [x] `go test ./...` (full suite, all packages pass)
- [x] `gatewaymigration.schema.json` regenerated via `go generate ./internal/manifest/...` and verified in sync (`TestGatewaySchemaInSync`)


ifeeneyconfluent:
Addressed the Copilot review finding on the lag-check fixture (\`cmd/migration/lagcheck/cmd_migration_lagcheck_test.go\`): the \`topicPatterns\` → \`topics\` swap in \`TestLagCheck_AlwaysWatchesEveryMirrorTopic\` still targeted the pre-\`spec.route\` four-space indentation, so \`strings.Replace\` silently no-op'd and the test never actually exercised an explicit topic selection.

Fixed in d68969be:
- Updated the replacement to match the current \`spec.route.topicGroup\` indentation.
- Added a \`require.NotEqual\` guard so a future fixture-format drift fails loudly instead of silently no-op'ing again.
- Added an assertion that the parsed manifest actually carries the explicit \`t1.order\` selection before checking \`cfg.Topics\` stays empty, so the test proves "explicit selection is ignored" rather than passing vacuously.

Verified by temporarily reintroducing the stale indentation locally — the new guard fails loudly as expected — then restoring the fix and re-running the full suite clean.

internal/manifest/schemagen/schemagen.go
```diff
@@ -156,11 +157,11 @@ func GenerateGateway() ([]byte, error) {
...
-		tgItem.Properties["topicPatterns"]:         "Topic selection as a list of anchored full-match regular expressions (RE2). Use ['.*'] to cut over every active mirror topic. At least one of topics or topicPatterns is required. Only consulted when topics is absent; only the match-all pattern is currently supported (union with topics is not yet implemented).",
-		tgItem.Properties["route"]:                 "The spec.routes[].name of the route to fence and switch over. Must exist in the initial CR and must not already be fenced.",
-		tgItem.Properties["targetStreamingDomain"]: "Name of a streaming domain already declared in the initial CR's spec.streamingDomains. kcp derives the bootstrap server id to bind the route to from that declaration in the live CR — it is not written here. Safe with no secret or auth change at cutover only because the route's security.cluster already carries pre-staged (\"redundant\") auth for this domain, which kcp proves at init.",
+		route.Properties["name"]:                  "The spec.routes[].name of the route to fence and switch over. Must exist in the initial CR and must not already be fenced.",
```
adrian-januzi: The new top-level `spec.route` object never gets a description in the generated schema, even though the old `spec.topicGroup` array it replaces had one, so this quietly drops it. The `describe()` map covers `route.name` (this line), `route.topicGroup` and its items, and `route.targetStreamingDomain`, but never `route` itself, and you can see it missing on the `route` object in the checked-in `gatewaymigration.schema.json`. Not a big deal, but anyone hovering `spec.route:` in a schema-aware editor now gets nothing telling them what that required section is for, and nothing tests object-level descriptions so it slipped through. Worth adding a `spec.Properties["route"]` entry back.

integration-tests/migration/testdata/gateway-migration.yaml.tmpl
```diff
@@ -46,7 +46,9 @@ spec:
     namespace: {{ .Namespace | yamlQuote }}
     kubeconfig: {{ .KubePath | yamlQuote }}
     cr-name: {{ .GatewayName | yamlQuote }}
-  topicGroup:
+  route:
+    name: {{ (index .FenceRoutes 0).Name | yamlQuote }}
```
adrian-januzi: Fixture-only, small one - this line and the `targetStreamingDomain` one down at 71 now pull the values out with `(index .FenceRoutes 0)`, which will blow the template render up with an index out of range if `FenceRoutes` is ever empty. The old form referenced `.Name` and `.SwitchoverDomainName` inside `{{ range .FenceRoutes }}`, so an empty slice just gave you empty output rather than a hard failure. Every scenario populates `FenceRoutes` today so nothing's actually broken, but it loses the tolerance the range had - might be worth guarding it or moving them back inside the range.

ifeeneyconfluent:
Also addressed the two wording nits from the same Copilot review (\`internal/manifest/gateway.go\` lines 44 and 88): \`Route\` was documented as "the dynamic route," which is misleading since the same field/type is shared by both static (AAO) and dynamic (TBM) migrations. Reworded to drop the "dynamic" qualifier in both the \`GatewaySpec.Route\` field comment and the \`Route\` type doc comment.

Fixed in a1fbbc2e.

internal/manifest/schemagen/schemagen.go
ifeeneyconfluent: Good catch — fixed in 4f68da9f. Added `route: "..."` back to the `describe()` map (mirroring the `Route` struct's own doc comment) and added an assertion in `TestGenerateGateway_PortsRetiredFlagGuidance` that `spec.route` itself carries a description, so an object-level description dropping again would fail the test.

integration-tests/migration/testdata/gateway-migration.yaml.tmpl
ifeeneyconfluent: Good catch — fixed in 4f68da9f. Wrapped both `(index .FenceRoutes 0)` lookups (`name` and `targetStreamingDomain`) in `{{ with .FenceRoutes }}...{{ end }}`, which restores the old range's tolerance: an empty slice now just omits those two lines instead of failing template execution with an index-out-of-range error. Added `TestRenderGatewayMigration_EmptyFenceRoutesDoesNotPanic`, which I verified fails against the unguarded template (reproducing the exact error you flagged) and passes with the guard.

adrian-januzi (approved):

ifeeneyconfluent:
@copilot review

adrian-januzi:
@claude review

adrian-januzi:
awww that would have been cool

